PR #154, opened 2014-05-13
use short names for submodule imports to ease imports out of mpl_toolkit...

gauteh:
...s

It is difficult to import the basemap module when it is not installed
along with the rest of the mpl_toolkits package. This change uses local
imports in the basemap package for submodules with potentially
conflicting global modules. This way it is easier to include the basemap
package when it is installed in a different location (e.g. a virtualenv
using system mpl_toolkits).

Python ensures that an import first looks in the containing package when
importing, this way it is not necessary to specify the full module name:

https://docs.python.org/2/tutorial/modules.html#intra-package-references

It addresses issues like:
http://stackoverflow.com/questions/18772557/how-can-one-locally-install-an-extension-to-an-existing-system-python-module


WeatherGod:
Sorry for taking so long to get to this. I plan to be much more attentive to this project now.

PR #181, opened 2015-02-23
style changes and removal of unused import (convert_path)

sbellem:


setup.py
```diff
@@ -95,8 +100,8 @@ def checkversion(GEOS_dir):
 # Specify all the required mpl data
 # create pyproj binary datum shift grid files.
 pathout =\
-os.path.join('lib',os.path.join('mpl_toolkits',os.path.join('basemap','data')))
-if sys.argv[1] not in ['sdist','clean']:
+    os.path.join('lib', os.path.join('mpl_toolkits', os.path.join('basemap', 'data')))
```
WeatherGod: If you could, please restructure this assignment so that it doesn't use the backslash?

WeatherGod: heck, just get rid of the gratuitous use of nested os.path.join()

setup.py
```diff
@@ -105,46 +110,46 @@ def checkversion(GEOS_dir):
...
-        fout = os.path.join(pathout,fout)
+        fout = os.path.join(pathout, fout)
         strg = '%s %s < %s' % (cmd, fout, f)
         sys.stdout.write('executing %s\n' % strg)
-        subprocess.call(strg,shell=True)
-datafiles = glob.glob(os.path.join(pathout,'*'))
-datafiles = [os.path.join('data',os.path.basename(f)) for f in datafiles]
-package_data = {'mpl_toolkits.basemap':datafiles}
+        subprocess.call(strg, shell=True)
+datafiles = glob.glob(os.path. join(pathout, '*'))
```
WeatherGod: this is incorrect "os.path.join" not "os.path. join"

WeatherGod:
Closing as this is significantly out of date now, and seems like the original author has lost interest.

PR #188, opened 2015-04-24
Allow basemap to be installed using pip, virtualenv and a requirements.txt file

markmuetz:
I ran into problems trying to install basemap using pip, virtualenv and a requirements.txt file. On installation, it tries to use setup.py to check the egg_info, and when it does this it throws an `ImportError` because numpy has not yet been installed (it is also in the requirements.txt file). 

Another project, [netcdf4-python](https://github.com/Unidata/netcdf4-python), had this problem and solved it by checking to see which command was being used with setup.py and only installing numpy if it was during installation: https://github.com/Unidata/netcdf4-python/commit/18f21a2ae2c21531f7d3bf8610cbc27cf2df2381. I copied the idea, with the main change being that setup.py still uses setup and Extensions from numpy during installation, as I wasn't sure if this was important and didn't want to break things for anyone.

The fix can be checked using a virtualenv. Create a virtualenv:

    virtualenv basemap_env
    cd basemap_env
    source bin/activate
    pip install -r requirements.txt

requirements.txt (works OK):

    numpy
    matplotlib
    git+https://github.com/markmuetz/basemap.git@pip_install#egg=basemap

requirements.txt (doesn't work)

    numpy
    matplotlib
    git+https://github.com/markmuetz/basemap.git@master#egg=basemap
    # or can just be:
    # basemap

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,10 +1,25 @@
-import sys, glob, os, numpy, subprocess
+import sys, glob, os, subprocess
+
 major, minor1, minor2, s, tmp = sys.version_info
 if major==2 and minor1<4 or major<2:
     raise SystemExit("""matplotlib and the basemap toolkit require Python 2.4 or later.""")
-from numpy.distutils.core  import setup, Extension
+
+from distutils.core import setup, Extension
+from distutils.dist import Distribution
 from distutils.util import convert_path
 from distutils import ccompiler, sysconfig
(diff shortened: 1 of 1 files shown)


tacaswell:
:+1: Seems fine to me.

@jswhit Should I feel free to merge changes like this?

PR #201, opened 2015-08-16
replaced ctlf with lf.

flamingbear:
Looks like someone checked in a windows edited version of this file at
some point (probably when it was in cvs).  Generally git will handle end
of line conversions for you.  This just removes the ^M from the end of each line and nothing else.


WeatherGod:
Good catch! Thanks for your PRs!

Issue #197, opened 2015-08-07
shiftdata exception when using latlon=True for scatter


benranck:
The exception;

```
m.scatter(x, y,50,marker='o',color='g',latlon=True, zorder=10)
  File Anaconda\lib\site-packages\mpl_toolkits\basemap\__init__.py", line 536, in with_transform
    x = self.shiftdata(x)
  File "Anaconda\lib\site-packages\mpl_toolkits\basemap\__init__.py", line 4777, in shiftdata
    if itemindex:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

There seems to be a mismatch when using latlon that shiftdata does not expect. This can be mitigated by getting the projection coordinates before applying the operation; example below.

```
from mpl_toolkits.basemap import Basemap
import numpy as np
import matplotlib.pyplot as plt

def valueError():
    m = Basemap(projection='robin',lon_0=115,resolution='c')
    m.drawcoastlines()
    m.fillcontinents(color='sienna',lake_color='aqua')
    m.drawmapboundary(fill_color='lightsteelblue')

    x = [-5.435, -6.660817, -119.2511944, 17.719833]
    y = [36.136, 4.746717, 39.4030278, -14.657583]
    m.scatter(x, y,50,marker='o',color='g',latlon=True, zorder=10)
    plt.show()

def works():
    m = Basemap(projection='robin',lon_0=115,resolution='c')
    m.drawcoastlines()
    m.fillcontinents(color='sienna',lake_color='aqua')
    m.drawmapboundary(fill_color='lightsteelblue')
    x = [-5.435, -6.660817, -119.2511944, 17.719833]
    y = [36.136, 4.746717, 39.4030278, -14.657583]

    x,y = m(x,y)
    m.scatter(x, y,50,marker='o',color='g',latlon=False, zorder=10)
    plt.show()

def main():
    works()
    valueError()

if __name__ == '__main__':
    main()
```

samskillman:
I just encountered the same issue.

guziy:
Actually sorting data by longitude like this seems to work, not, sure if this is a good fix... But maybe we could just sort longitudes in scatter if latlon is True??:

```python
from mpl_toolkits.basemap import Basemap
import numpy as np
import matplotlib.pyplot as plt


def valueError():
    m = Basemap(projection='robin',lon_0=115,resolution='c')
    m.drawcoastlines()
    m.fillcontinents(color='sienna',lake_color='aqua')
[...]

Issue #191, opened 2015-07-07
Problems with Basemap.scatter


bjornaa:
Hi,

I have stumbled over a bug in scatter plot. With latlon=True, it crashes with zero, one or two points.

Example code:

```python
import matplotlib.pyplot as plt
from mpl_toolkits.basemap import Basemap

bmap = Basemap(projection='merc',
               llcrnrlon=9, urcrnrlon=13,
               llcrnrlat=63, urcrnrlat=65,
               resolution='c')

lon = [10.0, 12.0]
lat = [64.0, 65.0]

bmap.scatter(lon, lat, latlon=True)
   
plt.show()
```

This crashes with the following Traceback:
```
Traceback (most recent call last):
  File "a.py", line 12, in <module>
    bmap.scatter(lon, lat, latlon=True)
  File "/opt/anaconda/lib/python2.7/site-packages/mpl_toolkits/basemap/__init__.py", line 536, in with_transform
    x = self.shiftdata(x)
  File "/opt/anaconda/lib/python2.7/site-packages/mpl_toolkits/basemap/__init__.py", line 4775, in shiftdata
    thresh = 360.-londiff_sort[-2]
IndexError: index -2 is out of bounds for axis 0 with size 1
```

I am using python 2.7.10, matplotlib 1.4.3, basemap 1.0.7 on a 64-bit Linux machine

pwolfram:
This may be a similar issue as #126

WeatherGod:
Once matplotlib 1.5 gets released, I am going to spend some time to fix up
basemap and get a new release of it out.

pwolfram:
@WeatherGod Ok, sounds good.  If you think this is something easy to fix I could try to fix it too.  Thanks!

WeatherGod:
I'd be more than happy to have more eyes on this codebase. I am not all
that familiar with it, and it is crufty, so I don't know how easy or hard
it would be to fix any particular issue. So, PRs on any aspect of Basemap
is welcomed.

One thing keeping me from more actively accepting patches is that TravisCI
isn't turned on for Basemap, so I don't know if a patch breaks things or
not. That should probably get fixed first.

My plan for Basemap is to put it into maintenance mode. I do not want to do
any new developments (although I won't refuse any PRs that adds features),
but these bug reports that have been accumulating shows that there is still
interest in this project and that it deserves to have things fixed at the
least. The intention is that cartopy will take over basemap, but cartopy
[...]

PR #202, opened 2015-08-21
Do not call shiftdata for scattter plots

guziy:
I've added a test for the scatter function and modified the decorator to eliminate the call to shiftdata for scatter in 1-d case, since I am not really sure why it is needed.... 

Addresses the issue #197.

Cheers

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -544,9 +544,9 @@ def _transform1d(plotfunc):
                     # shift longitudes to be in the range lon_0 +/- 180
                     lon_0 = self.projparams["lon_0"]
                     try:
-                        delta = (x - lon_0) % 360 + lon_0
+                        delta = (x - lon_0) % 360
                     except TypeError:
-                        delta = [(xx - lon_0) % 360 + lon_0 for xx in x]
+                        delta = [(xx - lon_0) % 360 for xx in x]
 
                     delta = np.asarray(delta)
                     delta[delta > 180] -= 360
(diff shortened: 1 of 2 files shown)


guziy:
The tests are added so it works with 1 and 2 points. (#191)

pelson:
@WeatherGod - did you have a case where this is not a good idea?

WeatherGod:
I don't think so. It was a few years ago that I reviewed this original
feature. I can't remember why we applied it to scatter plots. The only
thing I can think of is one is doing an image and scatter. One would want
to be able to have the domains for them match up, right?

guziy:
@WeatherGod  I'll try to generate a test case for this, to see if it really is a problem.

Cheers

guziy:
Apparently, shiftdata is needed for shifting when longitudes are further than +/- 180 from the lon_0:

https://github.com/guziy/PyNotebooks/blob/master/test_scatter_and_shiftdata.ipynb

I was supposing that it is done in the m(lon, lat) call....

One way to patch it would be a special shiftdata for scatter, which does not care about relations, but only changes coordinates to be within lon_0 - 180 and lon_0 + 180 range... Or change the __call__ method, however, I am not sure how it would impact other plotting methods.

Cheers

guziy:
The plot with shiftdata does not seem right when you look at the latitudes of the rightmost part...

guziy:
I was thinking about this a bit more, I think there should be a better way to solve this, not hacking only scatter and than maybe others... I'll try to understand shiftdata and fix it if possible...

Cheers

Issue #209, opened 2015-10-08
Problem projecting 2d arrays of coordinates..


guziy:
I know we have a lot of issues already, but this one is intriguing and maybe gives an opportunity to learn more about cython... 

Basemap works fine in the following situations (below lon2d and lat2d are 2D numpy arrays of latitudes and longitudes):

a)
```            
xx, yy = basemap(lons2d.T, lats2d.T)
```
b) 
```
xx, yy = basemap(lons2d.flatten(), lats2d.flatten())
```
c)  Looping through each point of the arrays and projecting works fine as well...

BUT

when I try to reproject the unchanged arrays
```python
 xx, yy = basemap(lons2d, lats2d)
```
I get the exception below:
```
In [4]: xx, yy = bmp(lons, lats)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-4-cd4fecf35ebb> in <module>()
----> 1 xx, yy = bmp(lons, lats)

/RESCUE/skynet3_rech1/huziy/python_virtualenvs/py3.4/lib/python3.4/site-packages/mpl_toolkits/basemap/__init__.py in __call__(self, x, y, inverse)
   1146             except TypeError:
   1147                 y = [_dg2rad*yy for yy in y]
-> 1148         xout,yout = self.projtran(x,y,inverse=inverse)
   1149         if self.celestial and inverse:
   1150             try:

/RESCUE/skynet3_rech1/huziy/python_virtualenvs/py3.4/lib/python3.4/site-packages/mpl_toolkits/basemap/proj.py in __call__(self, *args, **kw)
    284             outxy = self._proj4(xy, inverse=inverse)
    285         else:
--> 286             outx,outy = self._proj4(x, y, inverse=inverse)
    287         if inverse:
    288             if self.projection in ['merc','mill','gall']:

/RESCUE/skynet3_rech1/huziy/python_virtualenvs/py3.4/lib/python3.4/site-packages/mpl_toolkits/basemap/pyproj.py in __call__(self, *args, **kw)
    386             _proj.Proj._inv(self, inx, iny, radians=radians, errcheck=errcheck)
    387         else:
--> 388             _proj.Proj._fwd(self, inx, iny, radians=radians, errcheck=errcheck)
    389         # if inputs were lists, tuples or floats, convert back.
    390         outx = _convertback(xisfloat,xislist,xistuple,inx)

_proj.pyx in _proj.Proj._fwd (src/_proj.c:1571)()

[...]

PR #212, opened 2015-10-25
Projecting issue209

guziy:
To #209:

It seems that the problem is the order of the coordinate arrays ('F' instead of 'C') that proj or geos does not like...

I've added the tests that fail with the current version of the module and work with the fix. 

Cheers


WeatherGod:
Wow, that is some ancient code! Thanks for fixing!

PR #217, opened 2015-10-27
Fix issue #210

guziy:
Interpolate alpha similarly as color values...


guziy:
Warpimage plots with the new version of warpimage:
https://github.com/guziy/PyNotebooks/blob/master/basemap_issue_210.ipynb

PR #220, opened 2015-10-27
Travis ci

WeatherGod:
I have no clue if this will work...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,12 @@ install:
     pip install --upgrade pip
     pip install --upgrade setuptools
     pip install $NUMPY
-   - cd geos-3.3.3
-     export GEOS_DIR=$HOME/.local/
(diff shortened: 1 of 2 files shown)


WeatherGod:
@mdboom, @tacaswell , I don't think I have the necessary permissions with this group to add basemap to the approved list for TravisCI: https://travis-ci.org/matplotlib

tacaswell:
Turned on, next push should run.

tacaswell:
I merged this so that if the next thing that hits is a different PR it won't automatically fail.

QuLogic:
Failed to build, unfortunately. Apparently, you forgot to install matplotlib?

WeatherGod:
Ah, apparently I forgot that basemap isn't using setuptools and won't automatically install dependencies. I'll make a new PR and we can see if this works next time.

PR #223, opened 2015-10-28
Projecting issue209

guziy:
#209 Update to the fix to make it compatible with numpy 1.6


WeatherGod:
Thank you. By the way, it is preferred to do rebases rather than merges for updating your branches.

guziy:
Thanks, I'll try to rebase in future.

Cheers

2015-10-28 10:52 GMT-04:00 Benjamin Root <notifications@github.com>:

> Thank you. By the way, it is preferred to do rebases rather than merges
> for updating your branches.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/matplotlib/basemap/pull/223#issuecomment-151870603>.
>



-- 
Sasha

Issue #214, opened 2015-10-26
error plotting points that wrap around longitudinal map boundaries in plot() method of Basemap


paitor:
Hi 

Using Python 2.7.8 and Basemap 1.0.7 there seems to be a problem plotting points that wrap around longitudinal map boundaries in plot() method of Basemap if the points wrap back and forth. The problem can be avoided by sorting the points by longitude prior plotting (but this does of course not work if plotting a segment based on the points since sorting will screw up the order of the points along the segment)

The following code illustrates the problem by plotting the same three points sorted in different ways (first plot works)
```python
#! /usr/bin/env python

import matplotlib
matplotlib.use('TkAgg')
import matplotlib.pyplot as plt
from mpl_toolkits.basemap import Basemap

bm = Basemap(llcrnrlon=0,llcrnrlat=-80,urcrnrlon=360,urcrnrlat=80,projection='mill')
fig=plt.figure(figsize=(8,4.5))   

# figure with three points work
bm.plot([-30,-20,10], [10,10,10], "ro", latlon=True)
plt.show()

# figure with less than three points fail
bm.plot([-30,10,-20], [10,10,10], "ro", latlon=True)
plt.show()
```
and the error running the above code (after closing the first successful plot window) is
```
:~/SNSN/ALERT/seedlink> python TryBasemap.py 
Traceback (most recent call last):
  File "TryBasemap.py", line 16, in <module>
    bm.plot([-30,10,-20], [10,10,10], latlon=True)
  File "[...]/python2.7/site-packages/mpl_toolkits/basemap/__init__.py", line 536, in with_transform
    x = self.shiftdata(x)
  File "[...]/python2.7/site-packages/mpl_toolkits/basemap/__init__.py", line 4777, in shiftdata
    if itemindex:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

regP

WeatherGod:
Digging into this, once I fixed the "truthiness" problem and added your unit test, it revealed some more bugs under the hood.

This is gonna be a fun rabbit hole!

WeatherGod:
I have created #224 that should partly address the problem. At the very least, it won't error out anymore.

guziy:
Related to #197

gerritholl:
See also [question on Stack Overflow](http://stackoverflow.com/q/31839047/974555).

WeatherGod:
[...]

PR #224, opened 2015-10-28
Truthiness

WeatherGod:
Partial fix to the problem found in #214 . Still need to figure out the "wandering wrap" problem...


WeatherGod:
Hmm, I wonder if I have a slight behavior change here. Previously, if an `array([0])` result was assigned to itemindex, I think it would have skipped out of this if-statement. Now, it would go in, and the array would get completely rolled around.

WeatherGod:
superseded by #332.

PR #226, opened 2015-10-29
Fix URL / name of GSHHG coastline database in docs

rwogburn:
It's now GSHHG and has a different URL: https://www.soest.hawaii.edu/pwessel/gshhg/

### DIFF
--- a/doc/users/geography.rst
+++ b/doc/users/geography.rst
@@ -3,8 +3,8 @@
 Drawing a Map Background
 ========================
 
-Basemap includes the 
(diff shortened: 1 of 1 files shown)


WeatherGod:
Thanks for noticing. Has the data itself changed enough to warrant updating our packaged version?

rwogburn:
It looks as though the main difference is that GSHHG includes rivers and political boundaries.

WeatherGod:
Ok, so I am guessing that they just simply merged the GSSH and GMT data into one dataset, which are both used by Basemap anyway. So we can probably just limit this to a documentation change as you have here. Thanks!

rwogburn:
Thank you!

PR #227, opened 2015-10-31
changed dimension of concatenation in warpimage

### LINKED ISSUE #211: shadedrelief() fails because of changes in numpy.concatenate
In `basemap/__init__.py` in Line 4151 the array `self._bm_lons` is created with one dimension. In line 4169 it is concatenated to itself, but not with dimension 0 but with dimension 1.

Since numpy version 1.10 this is an error.

r5r3:
this commit fixes #211

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -4166,7 +4166,7 @@ class Basemap(object):
                 # stack grids side-by-side (in longitiudinal direction), so
                 # any range of longitudes may be plotted on a world map.
                 self._bm_lons = \
-                np.concatenate((self._bm_lons,self._bm_lons+360),1)
+                np.concatenate((self._bm_lons,self._bm_lons+360),0)
                 self._bm_rgba = \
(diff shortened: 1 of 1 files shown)


lib/mpl_toolkits/basemap/__init__.py
```diff
@@ -4166,7 +4166,7 @@ def warpimage(self,image="bluemarble",scale=None,**kwargs):
                 # stack grids side-by-side (in longitiudinal direction), so
                 # any range of longitudes may be plotted on a world map.
                 self._bm_lons = \
-                np.concatenate((self._bm_lons,self._bm_lons+360),1)
+                np.concatenate((self._bm_lons,self._bm_lons+360),0)
                 self._bm_rgba = \
                 np.concatenate((self._bm_rgba,self._bm_rgba),1)
```
WeatherGod: Does it need to be done down here, too? Would be nice to have a test, too.

r5r3:
No, `_bm_rgba` is the actual image. It is correctly concatenated along dimension 1.

WeatherGod:
Ok. I'll worry about the test when I convert the examples suite into a test suite (soonish...). Thank you!

PR #231, opened 2015-11-28
make pyshp optional

jdkloe:
here is my attempt to make pyshp optional. It introduces an environment variable PYSHP_dir that can be set to use an external copy of pyshp. If it is undefined the bundled version is used.
I could not find a "standard" method to do this in setup tools, so the changes in setup.py look somewhat like a workaround. It copies opt_lib/mpl_toolkits/basemap/shapefile.py to lib/mpl_toolkits/basemap/shapefile.py if the bundled version is to be used, then runs the setup function, and removes the file again after running this function.
Anyway, it seems to work correctly on my system, so if you think this is acceptable, you're welcome to merge in this pull request.


micahcochran:
I am glad that you are working on this.  I think this patch is a little more complicated than it should be.  

Here's what I'm thinking.  setup.py needs to be modified to make pyshp and pyproj packages dependencies, so setuptools needs to be added.  pyshp is really easy to install because it has no external dependencies. (pyproj has the C library PROJ.4 version 4.9 dependency.)

If you want to maintain backward compatibility for the internal shapefile.py you'd replace it with something like this:
```python
from shapefile import *
raise DeprecationWarning ("please import shapefile package directly (since version 1.0.8)")
```

My setuptools skills are not yet up to the task.

jdkloe:
Hi, thanks for looking at my pull request. I agree that it is somewhat ugly, and if a clean way is possible, then that should be preferred. I'm afraid I am not familiar enough with the differences between setuptools and distutils, at the moment to know how to do this cleanly. I'll try to study the details as soon as my time permits.

jenshnielsen:
Replaced by #234

PR #233, opened 2015-11-30
makes netCDF4 optional for examples/run_all.py

micahcochran:
If you do not have netCDF4 installed, some example tests will give ImportErrors.

This patch makes netCDF4 optional for run_all.py.  It imports netCDF4, if that package is not available, it removes tests that require netCDF4.

There may be some of the tests in the lines above the try-statement that could be put below the except-statement.  I did not look at that aspect.

### DIFF
--- a/examples/run_all.py
+++ b/examples/run_all.py
@@ -16,6 +16,7 @@ try:
     from netCDF4 import Dataset
 except ImportError:
     # remove tests requiring netCDF4
+    sys.stdout.write("Could not import netCDF4, skipping tests that require netCDF4.\n")
     test_files.remove('streamplot_demo.py')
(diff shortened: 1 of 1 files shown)


WeatherGod:
Could you include a message stating that some tests are being removed due to the netCDF4 package not being importable?

micahcochran:
Are you wanting me to print that message for the user?  Or are you wanting me to make this comment "# remove tests requiring netCDF4" a little more descriptive?

WeatherGod:
a message to the user so that they know that not all of the tests are being
executed. It can be a simple print function rather than using the warning
module.

micahcochran:
Thank you!

Issue #230, opened 2015-11-26
option to not use bundled libraries


jdkloe:
the new bundling policy of Fedora requires to ask upstream, to provide the option to not use the bundled libraries, but use system libraries in stead.
For basemap, currently geos already has this option, but also pyproj and pyshp are bundled. These last two do not have the option to use the system version of the library. Are there fundamental reasons why this is not provided? 
If not would you be willing to merge a pull request from my side that implements this?

WeatherGod:
I can only surmise that the reason for bundling pyshp and pyproj has to do
with historical issues with dependency handling. I would be more than happy
to accept a PR that makes this optional.

micahcochran:
I'm working on making pyproj and pyshp external dependencies.

In my opinion, it would make development cleaner because those libraries have entirely separate github development sites.  Reporting of issues with those code bases could be referred to those projects.  Many users will probably install the most recent version of the dependent packages.  Some may already have those packages installed.

WeatherGod:
Agreed. Let's aim to not bundle at all.

micahcochran:
[Here is what I've gotten so far.](https://github.com/micahcochran/basemap/tree/ext-pyshp-pyproj)

I still need to update setup.py.  distutils doesn't seem to understand install_requires, which is where dependencies are listed.  While running this setup.py I got this:
```
/usr/lib/python3.4/distutils/dist.py:260: UserWarning: Unknown distribution option: 'install_requires'
```
I will need to investigate if setuptools is the only way to do this and how numpy.distutils stack up.

WeatherGod:
setuptools introduced the "install_requires" feature, not distutils. This
stuff will need to be ported over to use setuptools if we want automatic
dependency resolution.

jdkloe:
So the question is if basemap should move to using setuptools instead of the simpler distutils. Are there any downsides to this?

WeatherGod:
@mdboom, do you have any thoughts on whether or not basemap should
transition to setuptools? The goals are to eliminate bundling. I wonder if
doing so would also possibly solve the whole namespace package issue.
Perhaps we might even consider transitioning away from namespace packaging
altogether?

micahcochran:
[...]

PR #234, opened 2015-12-01
Make pyshp and pyproj external dependencies

micahcochran:
This is in response to Issue #230.

This patch makes pyshp and pyproj packages external dependencies.  It most removes shapefile.py and pyproj.py.  It still maintains import locations for pyproj and shapefile, so code such as `from mpl_toolkits.basemap.shapefile import Reader` should still work.

I have NOT tested this.  The unitesting in test.py does not make use of shapefiles, and I'm unsure if it will catch a problems with pyproj.  There very will could have been something that I've missed.  For the moment, this might need to go in a separate branch.  

Installation will be a little harder due to the new dependencies.

src folder with PROJ4 has been left untouched.


.travis.yml
```diff
@@ -32,6 +32,7 @@ install:
     pip install --upgrade setuptools
     pip install $NUMPY
     pip install $MPL
+    pip install -r requirements.txt
   - |
     cd geos-3.3.3
     export GEOS_DIR=$HOME/.local/
```
WeatherGod: I wonder if the GEOS stuff can be eliminated? Wasn't only for pyproj? (I could be wrong, though).

micahcochran: This is failing because I removed the code from setup.py to compile _geoslib.c which is for GEOS library.  Oops.

At first glance, [shapely](http://toblerity.org/shapely/manual.html) looks like it could be replaced GEOS.  Also, that would put the burden of installing GEOS onto shapely package.

WeatherGod: Right, I suspect we don't need to compile geos anymore. So try taking that
out of the travis yaml. As for shapely, there are definite benefits to
updating the codebase to depend on that very nice package, but I think that
is orthogonal to the problem at hand.

micahcochran: I assume compiling GEOS isn't a huge issue.

Migrating to shapely should be done at a later point in time.

WeatherGod:
Hmm, looks like the import of pyproj is still getting messed up somehow. The `pyproj` that becomes available is essentially empty, causing the attribute errors Travis is seeing.

lib/mpl_toolkits/basemap/pyproj.py
```diff
@@ -1,868 +1,4 @@
-"""
-Cython wrapper to provide python interfaces to
-PROJ.4 (http://trac.osgeo.org/proj/) functions.
+# pyproj compatibility layer for externally hosted package
+# pyproj development occurs at https://github.com/jswhit/pyproj
```
WeatherGod: This file probably needs to be eliminated entirely. It is getting picked up by the pyproj imports the other files are doing because it is in the same directory.

WeatherGod:
Progress... but it looks like we have angered the py3k gods somehow. Nothing obvious to me at the moment.

micahcochran:
From my local testing I think it is pyproj.  I tested this on Python 3.4.3.  I installed pyproj 1.9.3 and 1.9.4 and both produced two errors on the same tests (`test_convert` and `test_results_should_be_same_for_c_and_f_order_arrays`).  

`test_results_should_be_same_for_c_and_f_order_arrays` is very new to test for #212, which is https://github.com/jswhit/pyproj/pull/29  . That is only in the development version of pyproj.  That is odd that it only impacts Python 3.

`test_convert` is very new test too.

I installed the development version (and added an import sys, which is a current PR), and there were no errors.

The development version of pyproj plays nicely with Python 3.x.

Here's where pyproj is at at the moment.   There is currently an [PR to add Travis-CI to pyproj](https://github.com/jswhit/pyproj/pull/33).  In a patch that I sent I forgot an import sys.
I will ask for the development version pyproj to be runnable, but a new pyproj release is likely needed.  Using the development version can cause unexpected problems like this.

WeatherGod:
Ok, so to help guide decision-making here, let's settle some questions. First, is there a work-around for this bug we are encountering? If not, which versions of pyproj are we going to have to blacklist?

Tangentially, this section of code is ancient and is causing a number of problems elsewhere. The reason for its existence was for back in the numeric/numarray days. Perhaps it makes sense to rip this stuff out (separately, of course)?

micahcochran:
I should revise my statement above:  both failing tests are from PR #223, which patch issue #209.  pyproj patched this same issue in https://github.com/jswhit/pyproj/pull/29 .

There could be a minimal version of pyproj, perhaps 1.9.3, for Python 2.x.  The minimum version pyproj for Python 3.x should probably the next release.

It is possible to skip the two (2) failing unit tests for Python 3 using current version of pyproj .  Doing something like this in test.py:
```python
import sys
PY3 = (sys.version_info[0] == 3)
  ...
@unittest.skipIf( PY3 and pyproj.__version__ <= "1.9.4", 
                  "TestProjectCoords tests skipped for pyproj version 1.9.4 for Python 3.x" )
class TestProjectCoords(TestCases):
   ...
``` 
That will cause it to skip the failing tests. IMO it is slightly unfair for an unpatched library to be expected pass new unit tests.

I see two courses of action.  (1.) Add the code above to skip the tests (my preference).  (2.) Wait for pyproj to get the development version working, install pyproj using the development version, but that means that basemap will probably should wait for a new version of pyproj to be released before it can have a new release.

I agree with getting rid of the old numeric/numarray code if it causes problems elsewhere.

micahcochran:
I think that I've got this working per the unit tests.

The syntax `from mpl_toolkit.basemap import pyproj` still works because there is an `import pyproj` statement in __init__.py.   `import mpl_toolkit.basemap.pyproj`  does not work.

Currently, `from mpl_toolkit.basemap import shapefile` does not work. The  could easily be added for some backward compatibility, if desired.  Let me know if you want me to make that change.

@WeatherGod  Thank you for all your help with this.  Sorry for gridlocking the project for about a week.

micahcochran:
I just thought of this; pyshp could easily be made an optional dependency.  It is required for `.readshapefile()` and `.drawcounties()`.  Only three of of the examples need it: fillstates.py, hurrtracks.py, and counties.py.

setup.py
```diff
@@ -1,7 +1,6 @@
 import sys, glob, os, subprocess
 
-major, minor1, minor2, s, tmp = sys.version_info
-if major==2 and minor1<4 or major<2:
+if sys.version_info < (2, 4):
```
jenshnielsen: I would be fine bumping the minimum version to at least 2.6 which is the oldest` one that we test with. Matplotlib has dropped support for 2.6 and earlier now

setup.py
```diff
@@ -161,7 +145,7 @@ def checkversion(GEOS_dir):
   packages          = packages,
   namespace_packages = namespace_packages,
   package_dir       = package_dirs,
-  ext_modules       = extensions,
+  ext_modules       = extensions,
```
jenshnielsen: Looks like this is unintentional

jenshnielsen:
:+1: looks good to me 

Making pyshp optional makes sense but that could just as easily be done in a different PR

WeatherGod:
I am -1 on making pyshp optional at this point. I know I have used it
myself when I didn't want the bulkiness of GDAL, and I have seen others use
it, too. I don't think there would be much to gain by making it optional,
versus the confusion it could cause.

I just remembered that there is a hidden optional dependency in basemap:
OWSLib for wmsimage(). Obviously, it isn't tested.

setup.py
```diff
@@ -81,57 +80,42 @@ def checkversion(GEOS_dir):
...
+    runtime_lib_dirs = geos_library_dirs
+
+extensions = [ Extension("_geoslib",['src/_geoslib.c'],
+                         library_dirs=geos_library_dirs,
+                         runtime_library_dirs=runtime_lib_dirs,
+                         include_dirs=geos_include_dirs,
+                         libraries=['geos_c']) ]
 
 # Specify all the required mpl data
 # create pyproj binary datum shift grid files.
```
jenshnielsen: Is this still true?

setup.py
micahcochran: It was accidentally I took it out and added it back it.  It was.

micahcochran:
Since pyshp is a python only library, it is so easy to install.  I don't see it being a problem to requiring it.  I thought I would bring it up.

setup.py
micahcochran: I was just trying to make the version requirements clearer without having all of those extra variables.  A minimum version of 2.6 is reasonable  Its last security fix, 2.6.9, was on October 29, 2013.

I'll modify that line to 2.6.

setup.py
micahcochran: No.  I'll remove that comment.

jenshnielsen:
:+1: Will wait for @WeatherGod to merge

WeatherGod:
Great work, @micahcochran! We will have to see about unbundling libgeos completely, but at least there is existing pathways for distributors to package it without the old geos codebase.

micahcochran:
@WeatherGod  Thanks for all your help.

jdkloe:
thanks for adopting the idea to remove these dependencies from the code base.

PR #236, opened 2015-12-08
remove unused c code from proj.4

jenshnielsen:
Following the merge of #234 we don't use the shipped proj.4 code so I suggest removing it. This removes all c code from src/ apart from _geos.c which is still used.


micahcochran:
:+1:

PR #237, opened 2015-12-08
Recythonize

jenshnielsen:
Regenerate _geos.c with an up to date version of Cython (0.23.4 vs 0.17) and fix some warning raised when cythonizing. Cython does not like more than one pointer defined pr line (I think this is due to the annotation features of cython)


WeatherGod:
Heh, didn't even realize that it was wrapped with cython. Thanks!

jenshnielsen:
In the slightly longer term we may consider doing this as a part of the build process? We could still ship the c file in the tar ball.

WeatherGod:
Perhaps. Ideally, I would like to externalize libgeos by using something
else like shapely or perhaps rasterio?

jenshnielsen:
Agreed that sounds sensible

Issue #235, opened 2015-12-04
Travis-CI:  Use  APT to install GEOS


micahcochran:
Install libgeos using apt would probably save almost 4 minutes of build time.  There is also a newer version of GEOS already packaged.

In .travis.yaml replace the compiling of libgeos with this:
```
addons:
  apt:
    packages:
    - libgeos-3.3.8
```

Here are the [docs for APT on Travis](https://docs.travis-ci.com/user/apt/).

I'm putting this here until I can get PR #234 working and merged.

jenshnielsen:
This sounds excellent. We may need to have the package whitelisted if not already done. See https://github.com/travis-ci/apt-package-whitelist

micahcochran:
There are currently 3 versions in the whitelist:  libgeos-3.2.2, libgeos-3.3.3, and libgeos-3.3.8.

jenshnielsen:
Great

WeatherGod:
Is it possible to use the APT version in some build and the bundled version
in other builds so that we can make sure that things work both ways?

micahcochran:
@WeatherGod It is possible.  That makes the configuration slightly more complicated.  I think the way to do that would be to have a GEOS_CURRENT flag that build version that compiles the most recent libgeos release, 3.5.0 for Python 2.7 & 3.5.  GEOS_DIR will have to be to set to the just compiled directory (it defaults to /usr/local).

The APT installed version will skip the compilation step when GEOS_CURRENT== True.

Maybe GEOS_CURRENT should be GEOS or GEOS_VERSION, I'm not that familiar with Travis-CI configuration.  I'm sure this approach can be refined.

micahcochran:
In order to accomplish this it might not be a bad idea if test.py has a function (software_versions)that prints out all of the software versions of installed modules.  It could be logged as informational an printed if the tests are ran verbosely (-v).

It might not be a bad way to diagnose future problems.

This could be in test.py or it could be a part of the standard library.

pelson:
FWIW, a *lot* of projects have had huge success using conda as the tool for package management on travis. That would allow full control of a GEOS / Proj.4 / ... version matrix, and is very quick to debug and diagnose. If you wanted to go down that route and have any questions, I'd be happy to help.

Cheers,

micahcochran:
[...]

PR #243, opened 2015-12-15
Add package_versions() and option to print ver. info in test.py

micahcochran:
package_versions() is a new function that returns a namedtuple of dependent package's version information.  It should also be useful to see if Travis is running the correct package versions.  I made this with Issue #235 in mind.  A --verbose option was added to test.py to display this version information.  Additionally, this new function could help with debugging future issues.

Example excerpt output from test.py:
>~/prj/basemap-master/lib/mpl_toolkits/basemap$ ~/prj/bmtest/bin/python3 test.py -v
Basemaps installed package versions
BasemapPackageVersions(Python='3.4.3+ (default, Oct 14 2015, 16:03:50) \n[GCC 5.2.1 20151010]', basemap='1.0.8', matplotlib='1.5.0', numpy='1.10.2', pyproj='1.9.4', pyshp='1.2.3', PROJ4=4.9, GEOS=b'3.5.0-CAPI-1.9.0 r4084', OWSLib='not installed', PIL='1.1.7', Pillow='3.0.0')

>test_convert (`__main__`.TestProjectCoords) ... skipped 'Test skipped in Python 3.x with pyproj version 1.9.4 and below.'

### DIFF
--- a/lib/mpl_toolkits/basemap/test.py
+++ b/lib/mpl_toolkits/basemap/test.py
@@ -224,7 +224,7 @@ if __name__ == '__main__':
     if VERBOSITY != None:
         pkg_vers = package_versions()
         print('Basemaps installed package versions:')
-        print('{}\n'.format(pkg_vers))
+        print('{0}\n'.format(pkg_vers))
 
     unittest.main()
 
--- a/lib/mpl_toolkits/basemap/test.py
+++ b/lib/mpl_toolkits/basemap/test.py
@@ -206,5 +206,25 @@ def test():
 
 
 if __name__ == '__main__':
+    # When called with the -v / --verbose commandline parameter, it will
+    # give package dependent version information:
+    #   $ python test.py --verbose
+
+    import argparse # standard module in Python >= 2.7 and >= 3.2
     import unittest
+
+    from mpl_toolkits.basemap.diagnostic import package_versions
+    
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--verbose',action='count')
+    args = parser.parse_args()
+    
+    # VERBOSITY will default to None if not a parameter
+    VERBOSITY = vars(args)['verbose']
+    if VERBOSITY != None:
+        pkg_vers = package_versions()
+        print('Basemaps installed package versions:')
+        print('{}\n'.format(pkg_vers))
+
     unittest.main()
+
--- a/lib/mpl_toolkits/basemap/test.py
+++ b/lib/mpl_toolkits/basemap/test.py
@@ -210,18 +210,12 @@ if __name__ == '__main__':
     # give package dependent version information:
     #   $ python test.py --verbose
 
-    import argparse # standard module in Python >= 2.7 and >= 3.2
+    import sys
     import unittest
 
     from mpl_toolkits.basemap.diagnostic import package_versions
     
-    parser = argparse.ArgumentParser()
-    parser.add_argument('-v', '--verbose',action='count')
-    args = parser.parse_args()
-    
-    # VERBOSITY will default to None if not a parameter
-    VERBOSITY = vars(args)['verbose']
-    if VERBOSITY != None:
+    if '--verbose' in sys.argv or '-v' in sys.argv:
         pkg_vers = package_versions()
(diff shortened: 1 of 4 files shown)


lib/mpl_toolkits/basemap/test.py
```diff
@@ -206,5 +206,25 @@ def test():
...
+
+    from mpl_toolkits.basemap.diagnostic import package_versions
+    
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--verbose',action='count')
+    args = parser.parse_args()
+    
+    # VERBOSITY will default to None if not a parameter
+    VERBOSITY = vars(args)['verbose']
+    if VERBOSITY != None:
```
WeatherGod: `!=` --> `is not`

lib/mpl_toolkits/basemap/test.py
```diff
@@ -206,5 +206,25 @@ def test():
...
+    #   $ python test.py --verbose
+
+    import argparse # standard module in Python >= 2.7 and >= 3.2
     import unittest
+
+    from mpl_toolkits.basemap.diagnostic import package_versions
+    
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-v', '--verbose',action='count')
+    args = parser.parse_args()
```
WeatherGod: The use of argparse is overkill here and would interfere with `unittest.main()`, which also processes command-line arguments. Instead, just import sys and inspect `argv` for "-v" and "--verbose".

micahcochran: Okay, I've submitted the patch for for this simpler approach.

WeatherGod:
This is looking good and it is a really good feature to have (been thinking about having something like this for matplotlib to make it easy for bug report triaging). One nitpit is that unlike all of the other version infos, Proj4's is a float, not a string. So, the Travis report for python2.6 (and 2.7? haven't checked) has proj4's version as 4.90000000003 or something silly like that. Don't know if I care enough, though.

micahcochran:
Python 2.6 for PROJ4 is showing a decimal rounding error:

BasemapPackageVersions(Python='2.6.9 (unknown, Feb  5 2015, 15:52:45) \n[GCC 4.6.3]', basemap='1.0.8', matplotlib='1.5.0', numpy='1.6.0', pyproj='1.9.4', pyshp='1.2.3', PROJ4=4.900000000000000**4**, GEOS='3.3.3-CAPI-1.7.4', OWSLib='not installed', PIL='not installed', Pillow='not installed')

The other Python versions show PROJ4=4.9

This need to be addressed. So, the current version of proj4,  4.9.2 should be 4.92.  The decimal version format is from pyproj.  That is inconsistent with all the other package versions.

This either needs to be rounded to 2 decimal places (so, the current version of proj4 '4.9.2', will be 4.92) OR reformatted into a string version number '4.9.2'. 

proj4's proj_api.h defines the version number like this:
```C
#define PJ_VERSION 492
```
pyproj simply divides PJ_VERSION by 100.  This encoding scheme may have to be changed if they go to a new minor release of 4.10.0.

micahcochran:
There is another way to get a string version out of proj4. It has a C function pj_get_release(), which returns a constant string, so the current version returns "Rel. 4.9.2, 08 September 2015".  pyproj currently doesn't expose this method.

WeatherGod:
Yeah, I definitely don't care enough to want add our own mechanism for
getting that information. If it is going to be painful, I am fine with it
as-is or just casting it as a rounded string.

micahcochran:
For now, I'm just going to accept that the proj4 version will be a decimal number based on what pyproj provides.  I'm not really anticipating for this function to be used for more than debugging and diagnostics.  I'm ready for merging if you are.

WeatherGod:
Travis logs for 2.6 shows that it still isn't rounded: `PROJ4=4.9000000000000004`

micahcochran:
I decided that having PROJ4 versions listed as the string '4.9.0' seems more natural.  It is also consistent with the way the rest of the software versions are shown.

WeatherGod:
Thank you

micahcochran:
Thanks

PR #244, opened 2015-12-16
stops ResourceWarnings for unclosed files

micahcochran:
Closes files bdatfile and bdatmetafile in the _readboundarydata function. 

This solves warnings like these on Python 3.x:
```
/.../lib/python3.4/site-packages/mpl_toolkits/basemap/__init__.py:1087: ResourceWarning: unclosed file <_io.TextIOWrapper name='/.../lib/python3.4/site-packages/mpl_toolkits/basemap/data/gshhsmeta_c.dat' mode='r' encoding='UTF-8'>
  self._readboundarydata('gshhs',as_polygons=True)
/.../python3.4/site-packages/mpl_toolkits/basemap/__init__.py:1095: ResourceWarning: unclosed file <_io.BufferedReader name='/.../lib/python3.4/site-packages/mpl_toolkits/basemap/data/gshhs_c.dat'>
  self._readboundarydata('gshhs',as_polygons=False)
```

This seems to be the simplest way to accomplish this.   ResourceWarnings are most of the warning raised when running test.py on Python 3.x.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -1460,6 +1460,8 @@ class Basemap(object):
                             if not as_polygons or len(b) > 4:
                                 polygons.append(list(zip(b[:,0],b[:,1])))
                                 polygon_types.append(typ)
+        bdatfile.close()
+        bdatmetafile.close()
         return polygons, polygon_types
 
     def _getmapboundary(self):


WeatherGod:
I was about to suggest to use with-statements instead, but then I noticed which method this was in. I wouldn't want to clutter our diff history just to indent that ungodly chunk of code.

One day, I'll rework that sucker. Even if it is just simply to say "I tackled that beast...". We will see what the output looks like shortly.

WeatherGod:
Much better! Now it looks like we have to deal with the non-integer indexing issue. I am sure there are much more instances of it than the warnings are now showing, though.

Merging!

micahcochran:
Thank you.

QuLogic:
Since the file open/close is basically around the entire function, the easiest way to avoid all that indent is to rename the function to something like `_readboundarydatafromfile` and have it take those two files as parameters, then create a new `_readboundarydata` that just opens the file in a `with` statement and only calls the renamed function.

PR #245, opened 2015-12-17
Use APT for libgeos on Travis

micahcochran:
This is for issue #235 . Using the APT binary versions of libgeos speed up Travis test times by eliminating some of the compiling.  Also, dashes were added to install steps to show the install timea for dependent packages.

In testing, the overall Travis test time is about the same.  It adds two additional tests builds.

The way this is setup APT installs these period.  setup.py does not have a way to force local libgeos install, which is probably not a bad thing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,18 +61,14 @@ matrix:
     # pip installs latest stable versions automatically
     - python: 2.7
       env:
-        - NUMPY=numpy>=1.10.0
-        - MPL=matplotlib>=1.5.0
         - BUILD_LIBGEOS=3.5.0
     - python: 3.5
       env: 
-        - NUMPY=numpy>=1.10.0
-        - MPL=matplotlib>=1.5.0
         - BUILD_LIBGEOS=3.5.0
     
     # "current development" environment
     # this test is diagnostic for development versions and is not required to pass
-    - python: "3.5-dev"
+    - python: "nightly"
       env:
         # these are python packages so prepend "git+" to git URL
         - NUMPY=git+https://github.com/numpy/numpy.git
@@ -82,7 +78,9 @@ matrix:
         - BUILD_LIBGEOS=https://github.com/libgeos/libgeos.git
         # or download latest stable
         # - BUILD_LIBGEOS=3.5.0
-
+        
+  allow_failures:
+    - python: "nightly"
 
 # before_install:
   # Install into our own pristine virtualenv
@@ -96,13 +94,13 @@ install:
   - pip install --upgrade setuptools
 
   # the development version of numpy requires Cython
-  - if [[ $TRAVIS_PYTHON_VERSION == '3.5-dev' ]]; then pip install --install-option="--no-cython-compile" Cython; fi
-  # workaround to get the specified numpy version to install on Python 2.7  
-  - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then pip install --upgrade $NUMPY; fi
+  - if [[ $TRAVIS_PYTHON_VERSION == 'nightly' ]]; then pip install --install-option="--no-cython-compile" Cython; fi
+
+  # workaround for "middle of the road" configuration to use pre-installed numpy, others "upgrade" numpy
+  - if [[ $TRAVIS_PYTHON_VERSION != '3.4' ]]; then pip install --upgrade $NUMPY; fi
   
-  - pip install $NUMPY
-  - pip install $MPL
-  - pip install $PYPROJ
+  - pip install --upgrade $MPL
+  - pip install --upgrade $PYPROJ
   - pip install -r requirements.txt
   
   # extra requirements to support Python 2.6  
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,14 +10,26 @@ language: python
 
 matrix:
   include:
+    # use libgeos binary from apt
     - python: 2.6
       env: NUMPY=numpy==1.6
     - python: 2.7
     - python: 3.3
     - python: 3.4
     - python: 3.5
+    
+    # compile internal libgeos
+    - python: 2.7
+      env: BUILD_INTERNAL_GEOS=true
+    - python: 3.5
+      env: BUILD_INTERNAL_GEOS=true
 #    - python: 2.7
 #      env: TEST_ARGS=--pep8
+
+addons:
+  apt:
+    packages:
+    - libgeos-3.3.8
 
 before_install:
   # Install into our own pristine virtualenv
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,15 +10,27 @@ language: python
...
     - python: 2.7
     - python: 3.3
     - python: 3.4
     - python: 3.5
+    
+    # compile internal libgeos
+    - python: 2.7
+      env: BUILD_INTERNAL_GEOS=true
+    - python: 3.5
+      env: BUILD_INTERNAL_GEOS=true
```
WeatherGod: Personally, I would have just picked two of the existing test runs and set them to use the internal geos. It'll reduce the amount of work that Travis has to do.

.travis.yml
```diff
@@ -27,24 +39,28 @@ before_install:
...
-    pip install --upgrade pip
-    pip install --upgrade setuptools
-    pip install $NUMPY
-    pip install $MPL
-    pip install -r requirements.txt
+  - pip install --upgrade pip
+  - pip install --upgrade setuptools
+
+  - pip install $NUMPY
+  - pip install $MPL
```
WeatherGod: looking at the timings for installing numpy and mpl, it makes me wish that binary wheels work for linux. Unfortunately, I think the conda route wouldn't really save us any time, either.

WeatherGod:
So, regardless of whether or not the libgeos was built locally or not, these tests will always use the system libgeos? Not sure if that was what you meant to say.

.travis.yml
micahcochran: There are APT binary versions for numpy 1.6.1 and matplotlib 1.1.1.  

I think I can install binary versions of freetype and libpng to speed up matplotlib compilation.

All the versions are old, but they are within minimum requirements.

WeatherGod: it is something to consider, but let's not get ahead of ourselves here. One thing I do want to make sure of is that we are testing against the minimum requirements as well as the current requirements, which is what the config is currently set up to do. If we were to install numpy and matplotlib from APT, then we lose that.

(Yes, I know we aren't testing against older matplotlibs. We should probably fix that)

micahcochran: I added freetype and libpng with APT in another branch.  The build was less than 10% faster, which probably has more to do with system load than any build time gains.

micahcochran: Spelling APT package names correctly is very important.  Installing via APT geos, freetype and libpng causes typical builds to be 2-3 minutes (per instance).  Compiling geos builds take about 6-7 minutes.  Where build times were taking 8-10 minutes.

micahcochran:
@WeatherGod disregard that.

Through trial and error, I did figure out how to apt install libgeos on some instances and not on others.  It made the configuration cumbersome and  it is not reflected in this PR.

I like the idea of having different versions of requirements to test against: Current, last release, and minimal requirements.  I think that should be another pull request to the travis,yaml.  The issue here is travis is kinda slow with the current configuration.

Do you want me to use perhaps Python 3.3 to test the compiling of the internal libGEOS?
Otherwise, I'm ready for a merge.

.travis.yml
jenshnielsen: You can use wheel caching on linux with travis. This builds the wheel the first time a package is used. You can also use the wheel house like we are doing in Matplotlib

micahcochran:
I almost forgot.  I removed virtualenv because [Travis already uses virtualenv by default](https://docs.travis-ci.com/user/languages/python#Travis-CI-Uses-Isolated-virtualenvs).

.travis.yml
jenshnielsen: In any case that should wait till another PR

micahcochran: I experimented with wheels a bit.  I didn't quite understand what I was doing wrong in the travis configuration.

jenshnielsen:
The Travis virtualenv comes with there version of numpy so not using our own virtual env means that we pick that one that Travis has unless we pass a specific version or pass the `--upgrade` flag

micahcochran:
With the current configuration for Python 2.6, pip uninstalls numpy 1.9.1 and then installs 1.6.2.

The current virtualenv version (currently in the repository) installs numpy 1.10.2.  Is that preferable?

I think specifying `env: NUMPY=numpy>=1.10,<1.11` on one of the versions will accomplish testing the latest & greatest.

What do you want to see before merging?

WeatherGod:
Let's have py27 and py35 be the "latest and greatest" environments, while py26 and py33 be the "minimum requirements" environments. In addition, the minimum requirements environments should use our build of libgeos. Any other environments should use the default provided by travis.

WeatherGod:
... and it looks like Travis had some sort of change and can't parse the yaml file correctly...

micahcochran:
@WeatherGod  My fault in copying and pasting...  I also forgot to make "3.5-dev" to allow failure.  I will do that in the next commit.

micahcochran:
The development version fails here with the unit test:
> ERROR: test_no_cyc2 (__main__.TestShiftGrid)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>  File "lib/mpl_toolkits/basemap/test.py", line 105, in test_no_cyc2
>    grid, lon = shiftgrid(lonin[len(lonin)/2], gridin, lonin, start=False)
> IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices

This is the same DeprecationWarning, but it is an IndexError in the development version.  The math here `lonin[len(lonin)/2]`  needs to use a double slash (//).  Why?  Well, that is the floor division operator.  If both numbers are integers, the resulting number will also be an integer.  

Example from Python 3.4.3
```
In [1]: 5/3
Out[1]: 1.6666666666666667

In [2]: 5//3
Out[2]: 1

In [3]: 5.0//3.0
Out[3]: 1.0
```

Should this IndexError/DeprecationWarning be address in this PR or another?

I assume there is an option enabled to make warning into errors in Python development version in order to keep warnings out of the mainline code.

micahcochran:
Notes about the current yaml configuration:

Downloads and compiles external version of libgeos from github.  Installs version that are "latest and greatest", "middle of the road", "minimum requirements", and "current development" environment.

The "middle of the road" Python 3.4.  Gets everything done in less than 3 minutes.

The "current development" environment is allowed to fail.  It is more of a diagnostic to address future possible incompatibilities.

For the "minimum requirements" environment, this took a bit of testing.  The earliest versions that I could get to work on Python 2.6 and 3.3 were matplotlib 1.2.0 and numpy 1.7.0.  Python 2.6 did work with earlier versions of numpy.  If desired, a travis test matrix could be built to test out all the possibilities to truly find the minimum.  See config for details.

Total build time is similar to before this, but this tests the boundaries a bit more.

There is more work that could be done, but I think this is a completely better system from the current.

.travis.yml
```diff
@@ -4,47 +4,138 @@ env:
...
+#        - NUMPY="numpy>=1.9.0,<1.10.0"  # use the default install of numpy
+      
+    # "latest and greatest" stable environments
+    # this uses the current stable versions
+    # pip installs latest stable versions automatically
+    - python: 2.7
+      env:
+        - NUMPY=numpy>=1.10.0
+        - MPL=matplotlib>=1.5.0
+        - BUILD_LIBGEOS=3.5.0
```
jenshnielsen: Rather than hardcoding the latest version which will for sure fall out off sync I would suggest these envs to pass `--upgrade` to pip that way we are sure that the latest version is installed.

.travis.yml
```diff
@@ -4,47 +4,138 @@ env:
...
-#    - python: 2.7
-#      env: TEST_ARGS=--pep8
+      env: 
+        - NUMPY=numpy>=1.10.0
+        - MPL=matplotlib>=1.5.0
+        - BUILD_LIBGEOS=3.5.0
+    
+    # "current development" environment
+    # this test is diagnostic for development versions and is not required to pass
+    - python: "3.5-dev"
```
jenshnielsen: I would suggest nightly in stead of 3.5-dev. Nightly is build from 3.6 master branch

micahcochran:
@jenshnielsen   I have incorporated your suggestions in the most recent commit.  The least amount of specifying a particular version makes it less likely that is configuration will have to be updated.  Python "nightly" is better.

WeatherGod:
This is looking real good now. We got one travis instance that is slow right now, but that is just luck of the draw, sometimes.

WeatherGod:
Ok, let's merge this and any additional improvements can be made in new PRs.

WeatherGod:
Great work!

micahcochran:
Thanks @WeatherGod

PR #246, opened 2015-12-19
fix for DeprecationWarning: using a non-integer

micahcochran:
Fixes these deprecation warnings from unitest in the TestShiftGrid class:
> test_cyc (__main__.TestShiftGrid) ... lib/mpl_toolkits/basemap/test.py:93: DeprecationWarning: using a non-integer number instead of an integer will result in an error in the future

~~This was done by changing the numpy arrays datatype from float to int.~~

### DIFF
--- a/lib/mpl_toolkits/basemap/test.py
+++ b/lib/mpl_toolkits/basemap/test.py
@@ -57,52 +57,52 @@ class TestShiftGrid(TestCase):
 
     def make_data_cyc(self):
         loncyc  =  np.array([0, 30, 60, 90, 120, 150, 180, 210, 240, 270, 300,\
-            330, 360], dtype=np.int)
+            330, 360],dtype=np.float)
         gridcyc = np.array([[0,  1,  2,  3,   4,   5,   6,   7,   8,   9,  10,\
-            11,   0]], dtype=np.int)
+            11,   0]],dtype=np.float)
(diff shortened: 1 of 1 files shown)


micahcochran:
Premature pull request.  Sorry.  Closing.

QuLogic:
FYI, you can update the branch and the PR will be updated; it's not necessary to close a WIP (unless you started with `master` branch because that's tricky.)

micahcochran:
Thanks @QuLogic .

It was actually an Warning caused by passing a float to the index of the numpy array.  In the development version of numpy, it causes an IndexError.

I am reopening this.

micahcochran:
I'm ready for this PR to be merged.  Is there a good reason not to merge?

WeatherGod:
I had completely forgotten about this. I can't remember what the hold-up was, but looking at the change, I don't think that this doesn't completely solve the problem. The result of `a//b` is still a float, so the result still needs to be cast to an integer, I think.

jenshnielsen:
@WeatherGod I don't think so 
```
3//2 = 1 
3.0//2 = 1.0
```
but len(lonin) should be an integer.

WeatherGod:
is that true in py3k? (I don't have a py3k environment on me at work)

jenshnielsen:
Yes that's the same in 3.5

WeatherGod:
Okie-doky, then.

micahcochran:
Thanks @WeatherGod and @jenshnielsen

PR #247, opened 2015-12-21
arcgisimage() wmsimage() should use print function

micahcochran:
Changes the code to use Python 3.x print functions instead of print statements.

Found this when I was playing with nose and nose2.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -4319,7 +4319,7 @@ format=png32&\
 f=image" %\
 (server,service,xmin,ymin,xmax,ymax,self.epsg,self.epsg,xpixels,ypixels,dpi)
         # print URL?
-        if verbose: print basemap_url
+        if verbose: print(basemap_url)
         # return AxesImage instance.
         return self.imshow(imread(urllib2.urlopen(basemap_url)),origin='upper')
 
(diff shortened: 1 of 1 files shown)


WeatherGod:
Just so you know, the entire codebase goes through 2to3 as part of the
installation process for python3 environments. Not ideal, but it works.

micahcochran:
Ahh that makes more sense now.

test.py for Travis isn't getting the 2to3 treatment.

I don't think there is any ideal way to support Python 2 and 3, just less bad ways.

micahcochran:
Should I close this as a useless PR?

jenshnielsen:
I think it's fine to convert the print statements to functions even if the code goes through 2to3. We should just add a `from __future__ import print_function` at the top too. 

It would be preferable to fully convert the code to use six or similar rather than 2to3 but thats a larger task

micahcochran:
2to3 will just strip out the `from __future import ...` statements.  If that is the future goal, then there is no time like the present to begin that change.

WeatherGod:
While I appreciate the enthusiasm, I would like to first direct that energy
towards building up the test suite. That test suite will be critical before
doing any sort of massive overhaul of the codebase.

micahcochran:
@WeatherGod I understand.  It is not a problem.

I do have some questions about expanding the unitests, but I will open an issue about that.

micahcochran:
Unneeded. Closing.

PR #248, opened 2015-12-23
Travis: cache pip and compiled libgeos binaries

micahcochran:
This enables caching of pip wheels and libgeos compiled binaries.  It should speed up build times for Travis-CI testing.

After the cache is built, most of the instances will build in less than 2 minutes.  The first time compiling it may take 6-15 minutes per instance.  On "nightly"  libGEOS will always build from source, not sure what pip. It should run about 10 minutes.

Times: [first build](https://travis-ci.org/micahcochran/basemap/builds/98557792) versus [already cached](https://travis-ci.org/micahcochran/basemap/builds/98556050)

Set "middle of the road" configuration to numpy=1.9.3.    A more flexible configuration caused too many problems in testing.  If you know a more maintainable way to do this, please submit a PR.

Cache stored is 670 MiB.

"nightly" will fail until PR #246 is merged.


.travis.yml
```diff
@@ -82,53 +87,58 @@ matrix:
   allow_failures:
     - python: "nightly"
 
-# before_install:
+before_install:
```
WeatherGod: why uncomment this if it isn't doing anything?

WeatherGod:
besides that one little nitpick, this looks ok to me.

.travis.yml
micahcochran: I keep flip flopping on switching virtualenv on and off.  I wasn't looking at it closely enough.  I'll fix it.

micahcochran:
Thank you!

WeatherGod:
No problem. By they way, could you email me directly? I don't have your email address.

PR #254, opened 2016-01-18
Correct Matplotlib documentation reference

jeixav:
The Basemap documentation for [pcolor](http://matplotlib.org/basemap/api/basemap_api.html#mpl_toolkits.basemap.Basemap.pcolor) refers to a Matplotlib function named tricolor. Such a function does not seem to exist; Matplotlib's [tripcolor](http://matplotlib.org/api/pyplot_api.html#matplotlib.pyplot.tripcolor) should probably be referenced instead.

This pull request changes matplotlib.pyplot.tricolor to tripcolor in the Basemap documentation


WeatherGod:
Thanks for catching that!

PR #258, opened 2016-01-30
Fix doc bug

jpli:
I think it's better to get the extra "plt.show()" removed since it dangles below the land-sea mask image created by the first "plt.show()" statement.


WeatherGod:
Thanks!

jpli:
You are welcome. Shall I open another pull request on gh-pages to reflect the change on the documentation site?

WeatherGod:
actually, no, I don't think that is where it needs to be done. I only
recently taken over this project, so I am still trying to figure out
exactly how deployment of the docs are done. I think it is all being
handled through ReadTheDocs, which I have zero experience with, nor do I
have an account on.

PR #259, opened 2016-02-09
Corrects the following error:

pick2510:
.7/site-packages/mpl_toolkits/basemap/__init__.py", line 4358, in wmsimage
    return self.imshow(imread(io.BytesIO(urllib2.urlopen(img.url).read()),
AttributeError: 'ResponseWrapper' object has no attribute 'url'

It might have something to do with an owslib update

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -4409,7 +4409,7 @@ f=image" %\
                          srs='EPSG:%s' % self.epsg, **kwargs)
         # return AxesImage instance.
         # this works for png and jpeg.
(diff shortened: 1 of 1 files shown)


WeatherGod:
Any idea how many versions owslib has supported this particular syntax?

pick2510:
Unfortunately not. I just tried to display a wms image with an all updated anaconda installation and debugged the wmsimage function. I didn’t build the package by hand to control the behaviour. But it solved my problem. 
I can give you more details how I called the function if that's useful for you

WeatherGod:
Ok, I did some digging into the history of owslib. The usage of calling ".read()" on the returned object dates back to at least 2009. I'd say that is sufficient backwards compatibility.

Issue #268, opened 2016-02-22
drawlsmask breaking with axis out of bounds error


AvlWx2014:
drawlsmask uses numpy.concatenate to stick 1-D lon arrays together on axis 1. Starting in NumPy 1.10 concatenate will raise an error if `axis` kwarg does not exist already.

micahcochran:
[Near the code in question](https://github.com/matplotlib/basemap/blob/master/lib/mpl_toolkits/basemap/__init__.py#L3960).

So you are saying that this line
```Python
lsmask_lons = np.concatenate((lsmask_lons,lsmask_lons[1:]+360),1)
```

Should be changed to this for NumPy 1.10:
```Python
lsmask_lons = np.concatenate((lsmask_lons,lsmask_lons[1:]+360),axis=1)
```

There are 4 other instances of np.concatenate in \__init\__.py

There currently isn't enough unit testing for geopandas to catch problems like this.

AvlWx2014:
NumPy concatenate will raise an IndexError, starting in NumPy 1.10.0 if the axis given does not exist i.e. lsmask_lons.ndim < 2 and `axis` = 1

WeatherGod:
I thought we already fixed these? Maybe not all of them needs changing:
https://github.com/matplotlib/basemap/pull/227

AvlWx2014:
I read #211 and it sounds like exactly the same issue. Looking at the code in __init__ in the master, it appears the issue still exists in drawlsmask. In my fork I'm going to use numpy.hstack which will successfully do what we're looking for, but works in old and new versions of NumPy.

PR #270, opened 2016-02-23
Update __init__.py

AvlWx2014:
Starting line 3960 within method drawlsmask I propose changing the use of numpy.concatenate to use numpy.hstack to achieve the stacking of the lsmask longitude array with itself. This is due to changes in numpy concatenate in NumPy version 1.10.0+ in which an IndexError will be raised if the axis of concatenation does not exist. Numpy.hstack will function the same way and persists in newer versions of NumPy.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -3960,10 +3960,14 @@ class Basemap(object):
             if cylproj:
                 # stack grids side-by-side (in longitiudinal direction), so
                 # any range of longitudes may be plotted on a world map.
+                # in versions of NumPy later than 1.10.0, concatenate will
+                # not stack these arrays as expected. If axis 1 is outside
+                # the dimensions of the array, concatenate will now raise
+                # an IndexError. Using hstack instead.
                 lsmask_lons = \
-                        np.concatenate((lsmask_lons,lsmask_lons[1:]+360),1)
+                        np.hstack((lsmask_lons,lsmask_lons[1:] + 360))
                 lsmask = \
-                        np.concatenate((lsmask,lsmask[:,1:]),1)
+                        np.hstack((lsmask,lsmask[:,1:]))
         else:
             if lakes: lsmask = np.where(lsmask==2,np.array(0,np.uint8),lsmask)
 


AvlWx2014:
This closes issue #268

WeatherGod:
Could you add a test for this problem, please? We are sorely in need for unit tests!

WeatherGod:
actually... now that I think about it, it really could only be tested via an image test, and we don't have the framework yet for it. Nevermind.

AvlWx2014:
I haven't written a test before but what I have is below. If you would like to incorporate it / it looks like it would work, I will create a branch and submit a pull request.

@skipIf(np.__version__ < "1.10.0", 
        "Test skipped if NumPy version is less than 1.10.x")
class TestLSMask(Basemap, unittest.TestCase):
	def make_map_object(self,**kwargs):
		""" Create a Basemap class instance """
		if PY3:
			m = super().__init__(**kwargs)
		else:
			m = super(TestLSMask,self).__init__(**kwargs)
		return m
	
	def test_running_drawlsmask_method_on_varying_map_objects(self):
		""" drawlsmask method should be callable without args supplied
		
		as kwargs have default values
		"""
		# Try some cases #
		# Case: all Basemap defaults
		m = self.make_map_object()
		self.assertRaise(IndexError,m.drawlsmask)
		
		# Case: lon_0=0, projection='cyl'
		m = self.make_map_object(lon_0=0)
		self.assertRaises(IndexError,m.drawlsmask)
		
		# Case: projection='merc'
		m = self.make_map_object(projection='merc')
		self.assertRaises(IndexError,m.drawlsmask)
		
		# Case: projection='npstere', lon_0=-100(center of US), 
		# boundinglat=40
		m = self.make_map_object(projection='npstere',
								lon_0=-100,
								boundinglat=40)
		self.assertRaises(IndexError,m.drawlsmask)

AvlWx2014:
`@skipIf(np.__version__ < "1.10.0", 
        "Test skipped if NumPy version is less than 1.10.x")
class TestLSMask(Basemap, unittest.TestCase):
	def make_map_object(self,**kwargs):
		""" Create a Basemap class instance """
		if PY3:
			m = super().__init__(**kwargs)
		else:
			m = super(TestLSMask,self).__init__(**kwargs)
		return m
	
	def test_running_drawlsmask_method_on_varying_map_objects(self):
		""" drawlsmask method should be callable without args supplied
		
		as kwargs have default values
		"""
		# Try some cases #
		# Case: all Basemap defaults
		m = self.make_map_object()
		self.assertRaise(IndexError,m.drawlsmask)
		
		# Case: lon_0=0, projection='cyl'
		m = self.make_map_object(lon_0=0)
		self.assertRaises(IndexError,m.drawlsmask)
		
		# Case: projection='merc'
		m = self.make_map_object(projection='merc')
		self.assertRaises(IndexError,m.drawlsmask)
		
		# Case: projection='npstere', lon_0=-100(center of US), 
		# boundinglat=40
		m = self.make_map_object(projection='npstere',
								lon_0=-100,
								boundinglat=40)
		self.assertRaises(IndexError,m.drawlsmask)`

micahcochran:
>```Python
@skipIf(np.version < "1.10.0", 
"Test skipped if NumPy version is less than 1.10.x")
```

We--I'm the biggest offender--need to stop using strings to compare library versions.  It works fine sometimes and other times it just fails.  For example:

```Python
In [68]: '1.8.0' < '1.9.0'
Out[68]: True

In [67]: '1.9.0' < '1.10.0'
Out[67]: False      ################ What??? ##############

In [69]: (1, 8, 0) < (1, 9, 0)
Out[69]: True

In [70]: (1, 9, 0) < (1, 10, 0)
Out[70]: True
```
Anyway,  I'm working on some code that will convert version strings into tuples. I hope to submit a PR soon.

WeatherGod:
Really, we should be using the LooseVersion class from distutils. We use it
in matplotlib all the time. This will help for situations where numpy was a
source build out of git and such.

AvlWx2014:
> Really, we should be using the LooseVersion class from distutils.

Pretty easy to incorporate. Requires an `from distutils.version import LooseVersion` at the top of the code, but that's not a big deal.

    @skipIf(LooseVersion(np.__version__) < LooseVersion("1.10.0")), 
            "Test skipped if NumPy version is less than 1.10.x")

micahcochran:
> Really, we should be using the LooseVersion class from distutils. 

I learned something new today.  Thanks!

PR #271, opened 2016-02-23
Update test.py

AvlWx2014:
Adds test of drawlsmask method to catch changes to NumPy functionality, which have broken drawlsmask recently. Recent changes to NumPy concatenate disallowed the concatenation of arrays on axes that don't exist. An IndexError is now raised when this is attempted. To test if drawlsmask functionality is working, the proposed addition is the TestLSMask class, which will fail the test if drawlsmask fails.

### DIFF
--- a/lib/mpl_toolkits/basemap/test.py
+++ b/lib/mpl_toolkits/basemap/test.py
@@ -190,8 +190,74 @@ class TestProjectCoords(TestCase):
         assert_almost_equal(xx1, xx2)
         assert_almost_equal(yy1, yy2)
 
-
-
+@skipIf(LooseVersion(np.__version__) < LooseVersion("1.10.0")), 
+        "Test skipped if NumPy version is less than 1.10.x")
+class TestLSMask(Basemap, unittest.TestCase):
+	def make_map_object(self,**kwargs):
+		""" Create a Basemap class instance """
+		if PY3:
+			# If Python 3, use new super() syntax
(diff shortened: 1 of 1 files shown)


lib/mpl_toolkits/basemap/test.py
```diff
@@ -190,8 +193,74 @@ def test_results_should_be_same_for_c_and_f_order_arrays(self):
...
+class TestLSMask(Basemap, unittest.TestCase):
+	def make_map_object(self,**kwargs):
+		""" Create a Basemap class instance """
+		if PY3:
+			# If Python 3, use new super() syntax
+			m = super().__init__(**kwargs)
+		else:
+			# Python 2 super syntax
+			m = super(TestLSMask,self).__init__(**kwargs)
+		return m
```
WeatherGod: I am not sure why you have this method. Why not just make a regular basemap object?

WeatherGod:
In the test code, please use 4 spaces for indents. I know it isn't followed correctly everywhere, but that'll get fixed at some point.

WeatherGod:
The error you are getting is because the unittests are running in headless mode, but the default backend is, in this case, tkagg. We need to have these tests run using the 'agg' backend. Add to the beginning of the test script:
```
import matplotlib
matplotlib.use('agg')
```

Also, because we don't have a proper test framework set up, this particular test needs to close the figure when cleaning up so that other they don't contaminate other (future) unit tests.

PR #273, opened 2016-02-29
fix #272

### LINKED ISSUE #272: nightshade UTC check
When trying to use the basemap `nightshade` method I noticed a bug in how basemap checks if the supplied `datetime` is in UTC. The relevant place in the codebase is [here](https://github.com/matplotlib/basemap/blob/8872dbc2bfd8682ad28d3b3ec7af553141be7c39/lib/mpl_toolkits/basemap/__init__.py#L4703)

```python
# make sure date is utc.
if date.utcoffset() is not None:
    raise ValueError('datetime instance must be UTC')
```

However datetime objects that are in UTC and are aware of it do not pass this check:

```python
In [1]: import datetime

In [2]: import pytz

In [3]: d = datetime.datetime.now()

In [4]: d
Out[4]: datetime.datetime(2016, 2, 29, 18, 21, 56, 388082)

In [5]: d.utcoffset() is not None
Out[5]: False

In [6]: d = d.replace(tzinfo=pytz.UTC)

In [7]: d
Out[7]: datetime.datetime(2016, 2, 29, 18, 21, 56, 388082, tzinfo=<UTC>)

In [8]: d.utcoffset() is not None
Out[8]: True

```

So even though my datetime has been made timezone aware (and converted from the original timezone to UTC), I have to set the timezone to `None` in order to pass the check for whether it is UTC... in which case it is strictly no longer UTC (just assumed to be so).

alpha-beta-soup:
Fixes problem with nightshade checking whether a datetime is in UTC; it was rejecting datetimes that have their `tzinfo` explicitly set to UTC. The check has now been corrected and simplified.

```python
In [37]: d = datetime.datetime.now()

In [38]: bool(d.utcoffset()) # time-zone naive
Out[38]: False

In [39]: d = d.replace(tzinfo=pytz.UTC)

In [40]: bool(d.utcoffset()) # utc
Out[41]: False

In [41]: eastern = pytz.timezone('US/Eastern')

In [42]: d = d.replace(tzinfo=eastern)

In [43]: bool(d.utcoffset()) # non-utc, non-zero offset
Out[43]: True
```

So therefore the check only needs to be `if date.utcoffset():`, which catches all non-zero offsets, and allows the naive case to continue passing.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -4700,9 +4700,9 @@ f=image" %\
         returns a matplotlib.contour.ContourSet instance.
         """
         from .solar import daynight_grid
-        # make sure date is utc.
-        if date.utcoffset() is not None:
-            raise ValueError('datetime instance must be UTC')
+        # make sure date is UTC, or naive with repect to time zones
+        if date.utcoffset():
+            raise ValueError('datetime instance must be UTC, not {}'.format(date.tzname()))
         # create grid of day=0, night=1
         lons,lats,daynight = daynight_grid(date,delta,self.lonmin,self.lonmax)
         x,y = self(lons,lats)


WeatherGod:
Thank you!

PR #279, opened 2016-03-14
Check geos version on Windows

ocefpaf:
This change ensures that `checkversion` will work on Linux and Windows.


WeatherGod:
Thank you!

PR #281, opened 2016-03-17
Add diagnostics functions: proj4 ver. & Inv. Hammer

micahcochran:
This is a refactor of the package_version function to move getting the proj4 version number into a new function, proj4_version.  check_proj_inv_hammer function is added for checking if the inverse of the hammer projection is available (which should be in the yet to be released proj.4 version 4.9.3).

check_proj_inv_hammer() and proj4_version() are intended to end up in setup.py, but may provide some use in the main basemap library.  Due to that reason, the imports are mostly self-contained for the function.

So, Proj.4 verion 4.9.3 has not been released, nor has the current development version of pyproj which has code that raises a RuntimeError when a projection's inverse function isn't available (pyproj 1.9.5.1 and before will segfault).  At the moment, I'm a little uncertain about putting these the check_proj_inv_hammer() into setup.py without one of the two libraries being released.  This effectively requires installation one of the development versions for the check_proj_inv_hammer function to pass the test, otherwise they risk a segfault.  This originates from discussion in PR #240.

Editing setup.py seems a little daunting because I'm a bit overwhelmed by what all setup.py can do/should do.  setup.py seems to work like a swiss army knife by doing a couple tasks well and other tasks not so well.

Here are my local test results of the functions, which where copied an pasted into ipython:
```Python
In [3]: proj4_version()
Out[3]: '4.9.2'

In [4]: package_versions()
Out[4]: BasemapPackageVersions(Python='2.7.4 (default, Apr  6 2013, 19:54:46) [M
SC v.1500 32 bit (Intel)]', basemap='1.0.8', matplotlib='1.5.1', numpy='1.10.4',
 pyproj='1.9.5.1', pyshp='1.2.3', PROJ4='4.9.2', GEOS='3.5.0-CAPI-1.9.0 r4084',
OWSLib='not installed', PIL='not installed', Pillow='not installed')

In [5]: check_proj_inv_hammer()
Out[5]: 'Unknown'

In [6]: check_proj_inv_hammer(True)
Out[6]: True
```

### DIFF
--- a/lib/mpl_toolkits/basemap/diagnostic.py
+++ b/lib/mpl_toolkits/basemap/diagnostic.py
@@ -2,6 +2,23 @@
 These are diagnostic and debugging functions for basemap.
 """
 
+def proj4_version():
+    """
+    Gives the proj.4 library's version number. (requires pyproj to be installed)
+
+    returns string, so proj.4 version 4.9.3 will return "4.9.3"
+    """
+    import pyproj
+    
+    # Get PROJ4 version in a floating point number
+    proj4_ver_num = pyproj.Proj(proj='latlong').proj_version
+    
+    # reformats floating point number into string (4.90 becomes '4.9.0')
+    # Exploits single number version numbers for proj4,
+    # This will need likely need to be change at some point as proj.4 version 4.10.0???
+    return '.'.join( str(int(proj4_ver_num*100)) )
+    
+    
 def package_versions():
     """
     Gives version information for dependent packages.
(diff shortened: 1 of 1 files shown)


WeatherGod:
Ok, this is an incremental improvement of the diagnostic functions. A nitpick: the double-negative for `segfault_protection_off` made me have to read the logic twice before I got it.

micahcochran:
I can understand the confusion.  In a few days, I'll do a commit to update the flag variable to just segfault_protection.Thanks for the quick feedback.

micahcochran:
I changed the parameter's name to `segfault_protection` and made another commit.

IPython console
```Python
In [3]: check_proj_inv_hammer(segfault_protection=True) # default
Out[3]: 'Unknown'

In [4]: check_proj_inv_hammer(segfault_protection=False)
Out[4]: True

In [5]: check_proj_inv_hammer()
Out[5]: 'Unknown'
```

If have an older version of Proj.4 or a version without the inverse hammer patch applied it will segfault Python.

micahcochran:
Thanks.

PR #284, opened 2016-04-07
Update version number code for pyproj

micahcochran:
Update code to reflect changes to pyproj API that simplify reporting version numbers.  Specifically, pyproj pull request 60,  https://github.com/jswhit/pyproj/pull/60

The geodesic library will report 'Unknown'.  See the "nightly" version to see the changes working and reporting a version (due to it pulling the development version of pyproj from github).

### DIFF
--- a/lib/mpl_toolkits/basemap/diagnostic.py
+++ b/lib/mpl_toolkits/basemap/diagnostic.py
@@ -9,14 +9,16 @@ def proj4_version():
     returns string, so proj.4 version 4.9.3 will return "4.9.3"
     """
     import pyproj
-    
-    # Get PROJ4 version in a floating point number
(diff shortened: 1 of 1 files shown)


WeatherGod:
Great, thanks for the updates. Should Cartopy get similar updates as well?

micahcochran:
Thank you.

Cartopy doesn't use pyproj, so it wouldn't be exactly the same.  Cartopy requires PROJ.4 (version 4.9.0 or above), which is they use their own Cython wrapper code [_crs.pyx](https://github.com/SciTools/cartopy/blob/master/lib/cartopy/_crs.pyx).  I could probably code something similar to what I did for pyproj.

PR #289, opened 2016-04-29
Add feet to drawmapscale

cellofellow:
I have a client that requires feet on the scale, so I added it.


micahcochran:
Could you add something to the message or `msg` about supporting feet?  (see https://github.com/cellofellow/basemap/blob/master/lib/mpl_toolkits/basemap/__init__.py#L4497)

cellofellow:
Sure, I'll update the doc

WeatherGod:
Yup, just update the msg variable to mention feet. Also, while you are mucking about in there, could you update the docstring to list the available names that can be used for `units`?

PR #290, opened 2016-05-05
The SuperCuts patch! Clipping done right.

### LINKED ISSUE #175: Imshow on non rectangular projections e.g. Robinson plots outside of map
Hi, 

You have already addressed this issue a few years ago [here](http://matplotlib.1069221.n5.nabble.com/Robinson-projection-imshow-out-of-bounds-and-different-latitude-limits-td15654.html) where you said it would be better to use pcolor. 

In my case I need the possibility of imshow to use a NxMx3 array as input to explicitly set the colors.

An example of the image I get and some code:

```
fig = plt.figure(figsize=(10, 18))
ax1 = plt.subplot(211)
m = Basemap(projection='robin',
            lon_0=0., ax=ax1)

# Compute lats and lons arrays
lons = (-180. + np.arange(720) / 2.) + 0.25
lats = (-90. + np.arange(360) / 2.) + 0.25

nx = int((m.xmax - m.xmin) / 25000.) + 1
ny = int((m.ymax - m.ymin) / 25000.) + 1

waterf_rp = m.transform_scalar(allf[:, :, 0], lons, lats, nx, ny)
radf_rp = m.transform_scalar(allf[:, :, 1], lons, lats, nx, ny)
tempf_rp = m.transform_scalar(allf[:, :, 2], lons, lats, nx, ny)

allf_rp = np.dstack((waterf_rp, radf_rp, tempf_rp))

# Plot Data
im = m.imshow(np.flipud(allf_rp))

# Add Grid Lines
m.drawparallels(np.arange(-90., 90., 30.), labels=[1, 0, 0, 0], fontsize=10)
m.drawmeridians(np.arange(-180., 180., 60.), labels=[0, 0, 0, 1], fontsize=10)

# Add Coastlines, States, and Country Boundaries
m.drawcoastlines()
m.drawstates()
m.drawcountries()
```
robinson problem example

I also tried to use the keywords masked and checkbounds in the transform_scalar method but they did not change the result.

It seems you already have a solution for the problem in the warpimage routine since the robinson projections works fine with. e.g. bluemarble ?

```
from mpl_toolkits.basemap import Basemap
import matplotlib.pyplot as plt
# setup Lambert Conformal basemap.
# set resolution=None to skip processing of boundary datasets.
m = Basemap(projection='robin',
            lon_0=0.)
m.bluemarble()
plt.show()
```
bluemarble_robinson

WeatherGod:
* Closes #175
* Clip just about all artists to map limbs,
  and clip the map limbs to the plot area

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -1668,7 +1668,7 @@ class Basemap(object):
                 limb.set_zorder(0)
             limb.set_edgecolor(color)
             limb.set_linewidth(linewidth)
-            limb.set_clip_on(False)
+            limb.set_clip_on(True)
             if zorder is not None:
                 limb.set_zorder(zorder)
         elif self.projection in _pseudocyl:  # elliptical region.
@@ -1702,7 +1702,7 @@ class Basemap(object):
             else:
                 limb.set_facecolor(fill_color)
                 limb.set_zorder(0)
-            limb.set_clip_on(False)
+            limb.set_clip_on(True)
             if zorder is not None:
                 limb.set_zorder(zorder)
         elif self.round:
@@ -1716,7 +1716,7 @@ class Basemap(object):
             else:
                 limb.set_facecolor(fill_color)
                 limb.set_zorder(0)
-            limb.set_clip_on(False)
+            limb.set_clip_on(True)
             if zorder is not None:
                 limb.set_zorder(zorder)
         else: # all other projections are rectangular.
@@ -1834,16 +1834,17 @@ class Basemap(object):
             npoly = npoly + 1
         # set axes limits to fit map region.
         self.set_axes_limits(ax=ax)
-        # clip continent polygons for round polar plots.
-        if self.round: polys,c = self._clipcircle(ax,polys)
+        # clip continent polygons to map limbs
+        polys,c = self._cliplimb(ax,polys)
         return polys
 
-    def _clipcircle(self,ax,coll):
-        c = Circle((0.5*(self.xmax+self.xmin),0.5*(self.ymax+self.ymin)),
-            radius=0.5*(self.xmax-self.xmin),fc='none')
+    def _cliplimb(self,ax,coll):
+        if not self._mapboundarydrawn:
+            return coll, None
+        c = self._mapboundarydrawn
         if c not in ax.patches:
             p = ax.add_patch(c)
-            p.set_clip_on(False)
+            #p.set_clip_on(False)
         try:
             coll.set_clip_path(c)
         except:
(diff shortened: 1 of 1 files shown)


WeatherGod:
attn @cpaulik

guziy:
@WeatherGod:  Nice one)) I was looking at it myself, but dropped because could not see why it was happening.

Cheers

WeatherGod:
Our unit tests wouldn't cover this, so I ran the "run_all.py" script in the examples directory. Not only did I check to make sure all of the examples ran correctly, I also would pan all of the interactive plots. With this change, not only does imshow() for robinson projection and others now work, but I can now confirm that all other artists are now clipped to the map, and to the plot area, even when panning/zooming. Of particular concern was the parallels and meridians, the coastlines and other such items, which all works fine now.

Technically speaking, this is a behavior change, as the plotting area was often not used for clipping. However, it was very inconsistent whether or not an artist was clipped or not, so I considered that a bug and made *everything* (except ticklines and tick labels) clipped to the plotting area.

sebhahn:
I just quickly tested your PR and something seems to be wrong with the projection. However, let me check if this is a problem on my side, since I've no general minimal example at hand.

test

sebhahn:
ok just figured out that it was my fault. I had to use `m.transform_scalar()`. Before, I used the image directly in `imshow()` and not the transformed one. Thanks again for fixing this issue!

test2

WeatherGod:
Thanks for confirming the fix! I am going to go ahead and merge this in.

If all goes well (personal schedule-wise), I hope to be cutting a release candidate either this weekend or next weekend. An additional goal is to have the release candidate available on pypi.

sebhahn:
@WeatherGod, not sure if this is related to this, but is there any option to set the thickness of the map boundary, it looks like the top/bottom line are much thinner than the left/right one. Looking closely to the line at the equator makes it also a little bit thinner. I know this is just a facial issue, but maybe you know how to control this?

WeatherGod:
Actually, this is an inadvertent side-effect of this recent change. The map limbs are now clipped to the plotting area, which solves a lot of other problems. However, because of the way the limbs are drawn, they go all the way to the edge of the plot area, and get "trimmed" by the clipping.

I am not exactly sure how we should address this. I'll have to think about this one.

sebhahn:
Yes, it seems like the lines are trimmed and therefore appear less thick at the top/bottom and the equator. A buffer zone around the lines/clipping area may help?

WeatherGod:
Easier said than done, I think. The problem is that the map limbs are done in data coordinates. So, defining a buffer zone around that in points is tricky. Also, I am not sure if there might be other implications as well (thinking of rotpol and such).

WeatherGod:
Hmmm, think I just came across another possible side-effect of this set of changes. It apparently has been assumed that set_axes_limits() would be the last method called in each of the plotting methods, so if one was in interactive mode, it would trigger a draw. However, now, that draw would happen before clipping. This might not be terrible, though, with the changes in the interactive mode for matplotlib v1.5, which should trigger another redraw right after the clipping, I think.

PR #291, opened 2016-05-05
Limb props

### LINKED ISSUE #160: Limb (map boundary) color is not set if round=True
In `drawmapboundary` the line/edge color of the limb is set in all cases except in the `round=True` case. This should not be.

The following is probably the problematic line:
```python
            limb = Circle((0.5*(self.xmax+self.xmin),0.5*(self.ymax+self.ymin)),
                    radius=0.5*(self.xmax-self.xmin),fc='none')
```

WeatherGod:
Consistent handling of limb properties. Closes #160. Depends (sort of) on #290.


WeatherGod:
Failure is unrelated (the bad pyparsing release). Merging.

Issue #299, opened 2016-05-25
Can't add scatter with 2 points in array


nathanscully:
Run into an issue where if the lng/lat array of coords provided to m.scatter(...) is 2 or less I get the following error:
```
Traceback (most recent call last):
  File "basemap_scatter.py", line 55, in <module>
    m.scatter(points2[0],points2[1], 500, c='blue',alpha=1, zorder=0.6, latlon=True)
  File "~/anaconda/lib/python3.5/site-packages/mpl_toolkits/basemap/__init__.py", line 536, in with_transform
    x = self.shiftdata(x)
  File "~/anaconda/lib/python3.5/site-packages/mpl_toolkits/basemap/__init__.py", line 4775, in shiftdata
    thresh = 360.-londiff_sort[-2]
IndexError: index -2 is out of bounds for axis 0 with size 1
```

single coords work and 3+ appears to work.

Example showing working/not working combinations I have tried. 

```
import matplotlib.pyplot as plt
from mpl_toolkits.basemap import Basemap
from shapely.ops import transform
from shapely.geometry import Polygon, Point, MultiPolygon
from descartes import PolygonPatch

#setup bounds
minx, miny, maxx, maxy = [144.699997,-37.882599,144.7813598,-37.753114]

#width height
w, h = maxx - minx, maxy - miny

plt.figure(figsize=(18,12))
m = Basemap(
    projection='merc',
    ellps = 'WGS84',
    llcrnrlon=minx - 3 * w,
    llcrnrlat=miny - 3 * h,
    urcrnrlon=maxx + 3 * w,
    urcrnrlat=maxy + 3 * h,
    resolution='h')
m.drawcoastlines()
m.drawmapboundary(fill_color='lightskyblue')
m.fillcontinents(color='silver',lake_color='lightskyblue', zorder=0.1)


point1 = [144.687,-37.751]
point2 = [144.793,-37.881]
point3 = [144.695,-37.742]


# =============================== Scatter tests ====== #

#single point - WORKS
#m.scatter(point1[0],point1[1], 500, c='blue',alpha=1, zorder=0.6, latlon=True)

##2 single points - WORKS
# m.scatter(point1[0],point1[1], 500, c='blue',alpha=1, zorder=0.6, latlon=True)
# m.scatter(point2[0],point2[1], 500, c='blue',alpha=1, zorder=0.6, latlon=True)

#Array of 3 elements - WORKS
# points3 = [list(i) for i in zip(point1,point2,point3)]
[...]

PR #300, opened 2016-05-26
Make shiftdata work with less than 2 points, add a test for this from the issue #299

guziy:
I tried to fix the 1 or 2 points problem in a more correct manner by modifying shiftdata. I've added tests from #299.. Please verify if anything else should be done..


guziy:
I forgot to include the test from the issue. It's on my home computer, I'll commit it in the evening if necessary.

guziy:
Actually, I have just tried the tests from #299 with the current master, i.e. not using this pull request, and it works, so I guess it was already fixed))

It seems to be a good idea to check an issue against the latest master before trying to fix it))

Cheers

PR #301, opened 2016-05-30
Add alpha blending to wmsimage

pablo-benito:
Add alpha argument to wmsimage(), so the image downloaded from the WMS server blends with the basemap


WeatherGod:
Thanks for the feature!

PR #305, opened 2016-06-24
Fix print-command for python3

uvchik:
The import does not work on python3 due to incompatible print commands in the init-file.

```python
>>> from mpl_toolkits.basemap import Basemap
  File "<stdin>", line 1, in <module>
  File "/home/uwe/rli-lokal/python_packages/basemap/lib/mpl_toolkits/basemap/__init__.py", line 4292
    if verbose: print basemap_url
                                ^
SyntaxError: Missing parentheses in call to 'print'
```

This PR just adds a few parenthesis to avoid these errors.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -4289,7 +4289,7 @@ format=png32&\
 f=image" %\
 (server,service,xmin,ymin,xmax,ymax,self.epsg,self.epsg,xpixels,ypixels,dpi)
         # print URL?
-        if verbose: print basemap_url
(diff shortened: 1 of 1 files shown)


WeatherGod:
basemap uses 2to3 upon install. Your install must have originally been done using python2. Use conda environments if you want to switch between multiple python versions.

uvchik:
I got sometimes strange ImportErrors using my basemap installation (1.0.7) so I decided to clone the repository and install it via ```pip3 install -e /path/to/repository/basemap```. Using my fix everything works better for me than before. No ImportErrors anymore :smile: So I will use my fix locally until version 1.0.8 is released.

Thanks anyway.

PR #308, opened 2016-07-08
BUG: use integer division where appropriate

efiring:
There are a few places where it is obvious that integer division is needed, and the 2to3 tool can't take care of them.  I tripped over one via a warning from numpy about non-integer indexing, and scanned `__init__.py` to find others.


WeatherGod:
Thanks! I was meaning to fix those up!

PR #312, opened 2016-07-21
Fix tab in drawlsmask

nparley:
When calling drawlsmask for a second time `_readlsmask` will not reload the data. However, `cylproj` was still be calculated and because `lsmask_lons` could be `None`  this will cause an error. This code does not need to be run if `self.lsmask` is not `None` so it's most likely a tab error and it should belong to the `if self.lsmask is None:` block.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -3901,11 +3901,11 @@ class Basemap(object):
         # look for axes instance (as keyword, an instance variable
         # or from plt.gca().
         ax = kwargs.pop('ax', None) or self._check_ax()
-        # Clear saved lsmask is new lsmask is passed
+        # Clear saved lsmask if new lsmask is passed
         if lsmask is not None or lsmask_lons is not None \
                 or lsmask_lats is not None:
             # Make sure passed lsmask is not the same as cached mask
-            if lsmask != self.lsmask:
+            if lsmask is not self.lsmask:
                 self.lsmask = None
         # if lsmask,lsmask_lons,lsmask_lats keywords not given,
         # read default land-sea mask in from file.
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -3901,6 +3901,9 @@ class Basemap(object):
         # look for axes instance (as keyword, an instance variable
         # or from plt.gca().
         ax = kwargs.pop('ax', None) or self._check_ax()
+        # Clear saved lsmask is new lsmask is passed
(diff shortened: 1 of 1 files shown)


WeatherGod:
Hmmm, this logic is still wrong, though. If I am reading this right, if someone passes in a new lsmask after already calling this function, then it doesn't get used, and it doesn't get adjusted for cyclic projections. Makes me wonder if no one passes in their own lsmask?

I would imagine that the correct way to deal with this is to set `self.lsmask` to `None` at the start of the function if a lsmask is passed in (and that lsmask is determined to be different from the current one). Then the rest of the logic is simplified.

nparley:
Maybe adding this in?:

```python
# Clear saved lsmask is new lsmask is passed
if lsmask is not None or lsmask_lons is not None or lsmask_lats is not None:
  self.lsmask = None
```

WeatherGod:
yeah, that should do the trick along with your current set of changes

lib/mpl_toolkits/basemap/__init__.py
```diff
@@ -3901,6 +3901,9 @@ def drawlsmask(self,land_color="0.8",ocean_color="w",lsmask=None,
         # look for axes instance (as keyword, an instance variable
         # or from plt.gca().
         ax = kwargs.pop('ax', None) or self._check_ax()
+        # Clear saved lsmask is new lsmask is passed
+        if lsmask is not None or lsmask_lons is not None or lsmask_lats is not None:
+            self.lsmask = None
```
WeatherGod: I might do one final little check. I could see adding a check to see if the passed in `lsmask` is the same as the cached `lsmask`.

nparley: Good point

lib/mpl_toolkits/basemap/__init__.py
```diff
@@ -3901,6 +3901,12 @@ def drawlsmask(self,land_color="0.8",ocean_color="w",lsmask=None,
         # look for axes instance (as keyword, an instance variable
         # or from plt.gca().
         ax = kwargs.pop('ax', None) or self._check_ax()
+        # Clear saved lsmask is new lsmask is passed
+        if lsmask is not None or lsmask_lons is not None \
+                or lsmask_lats is not None:
+            # Make sure passed lsmask is not the same as cached mask
+            if lsmask != self.lsmask:
```
WeatherGod: Don't use equality. Use `is not` instead when comparing objects.
Also, slight typo in the comments above "lsmask is new" --> "lsmask if new"

WeatherGod:
Thanks for your fix!

PR #311, opened 2016-07-18
Update gshhs/gshhg data to GMT 5.x and GSHHG 2.3.5

joferkington:
This PR updates the coastline, river, and political boundary data to the latest version of GSHHS/GSHHG.

A few notes:  

Based on the metadata that was stored, it's obvious that the previous data was from the shapefile versions of GSHHS and not the "raw" GMT version.  Therefore, I've generated this version from the shapefiles as well, and changed the README to note that that's the preferred route.  However, I've also fixed the GMT scripts so that GMT can now be used to generate the data as well. 

One side effect of this is that _GMT is no longer required to update the data_ in most cases.  Only the shapefiles (a separate download) are necessary.  GMT is still required to update the landmasks, but they do not change frequently at the resolution they're stored at and should only need very infrequent updates.

@WeatherGod - Finally managed to get all of this working.  I've tried to keep it to as minimal edits as possible, so the scripts are still a bit crufty, but they work now.

PR #320, opened 2016-09-23
Update data

WeatherGod:
A remade version of #311 using GSHHG 2.3.6. Also drops out all of that PR's commits that changed the data in order to keep the repo size down. Also fixes a bug discovered during testing of that PR.

Huge thanks to @joferkington for his critical work in #311, without which this would not have been possible!


WeatherGod:
Hmm, just noticed that I left out the changes to some of the other data like lsmask.

WeatherGod:
There we go, now both PRs modify all the same files. I am not concerned that the lsmask data came from v2.3.5 of the GSHHG data given that the only difference between v2.3.5 and v2.3.6 is fixing some errors with the antartica data, which I believe Joe already took into account.

WeatherGod:
The tests have passed. Since this PR was originally Joe's with just a tiny little fix, I am going to go ahead with a self-merge.

PR #328, opened 2016-11-27
Convert hold-related functions to attribute assignments.

efiring:
This is a minimal change so that the deprecation of "hold"
in mpl 2.x won't trigger warnings in Basemap from routine
plotting calls.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -3225,16 +3225,16 @@ class Basemap(object):
         """
         ax, plt = self._ax_plt_from_kw(kwargs)
         # allow callers to override the hold state by passing hold=True|False
-        b = ax.ishold()
+        b = ax._hold
         h = kwargs.pop('hold',None)
         if h is not None:
-            ax.hold(h)
+            ax._hold = h
         try:
             ret =  ax.scatter(*args, **kwargs)
         except:
-            ax.hold(b)
+            ax._hold = b
             raise
-        ax.hold(b)
+        ax._hold = b
         # reset current active image (only if pyplot is imported).
         if plt:
             plt.sci(ret)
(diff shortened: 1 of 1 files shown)


efiring:
Refer to https://github.com/matplotlib/matplotlib/pull/7516 in mpl.

WeatherGod:
sorry, I missed this earlier this month. This makes sense to do, but perhaps we should be doing `getattr(ax, '_hold', False)` to be future-proof?

WeatherGod:
or True, None... whateer...

efiring:
@WeatherGod: good idea to future-proof.  I think I have done that now.  Maybe I should have used a context-manager, but oh, well.

lib/mpl_toolkits/basemap/__init__.py
```diff
@@ -3206,6 +3206,17 @@ def set_axes_limits(self,ax=None):
             import matplotlib.pyplot as plt
             plt.draw_if_interactive()
 
+
+    def _save_use_hold(self, ax, kwargs):
+        h = kwargs.pop('hold', None)
+        if hasattr(ax, '_hold'):
+            self._tmp_hold = ax._hold
+            ax._hold = h
```
WeatherGod: This assignment only happens if the axes already has a `_hold` attribute. So, this wouldn't work if someone was setting the hold initially.

lib/mpl_toolkits/basemap/__init__.py
```diff
@@ -3206,6 +3206,17 @@ def set_axes_limits(self,ax=None):
...
+
+    def _restore_hold(self, ax):
+        if hasattr(ax, '_hold'):
```
WeatherGod: I would check for `hasattr(self, '_tmp_hold')` instead, and then del the attribute when done

lib/mpl_toolkits/basemap/__init__.py
efiring: My thinking here is that presently (with 2.x), the _hold attribute is always present.  It is set in the AxesBase initializer.  Some time in the future it won't be, and at that point Basemap doesn't need to obey it, and shouldn't set it at all.

WeatherGod: ah, yeah, that does make sense then. And it prevents basemap from adding the attribute itself if it isn't there already.

PR #331, opened 2016-12-30
Add keyword arguments `linewidth' and `linecolor' to drawmapscale.

mlegner:
I needed to change some properties of the mapscale recently and would like to suggest adding keyword arguments controlling the linewidth and color of the mapscale.


WeatherGod:
looks good to me. Waiting for Travis to complete, but I doubt it'll find any problems.

WeatherGod:
Thanks for the feature!

PR #332, opened 2017-01-02
Shiftdata fix

guziy:
I've tried to avoid the portion of shiftdata trying to do the wraparound fix for the scatter. For this I've added a keyword argument to the shiftdata and then execute the wraparound fix only if the argument is True. It is True by default but is set to Falsewhen the plotfunc is scatter.

That should close the #197, but to be sure I would like to wait for the failing example from @kiwi937.

Refactor: move masking of the output longitudes and data arrays for 1d and 2d cases to a single place, since they are the same.

Cheers

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -4811,13 +4811,7 @@ f=image" %\
                         datain_save[:,1:] = datain
                         datain_save[:,0] = datain[:,-1]
                         datain = datain_save
-                # mask points outside
-                # map region so they don't wrap back in the domain.
-                mask = np.logical_or(lonsin<lon_0-180,lonsin>lon_0+180)
-                lonsin = np.where(mask,1.e30,lonsin)
-                if datain is not None and mask.any():
-                    # superimpose on existing mask
-                    datain = ma.masked_where(mask, datain)
+
         # 1-d data.
         elif lonsin.ndim == 1:
             nlons = len(lonsin)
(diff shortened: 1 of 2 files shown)


WeatherGod:
I'd feel better about this if there were some unit tests added. Also, what was the reason for excluding scatter() from shiftdata? Perhaps it would be good to add some more explanation into docstrings and/or comments.

Also, and perhaps this is because the docstring for shiftdata isn't all that great, but an initial read of the doc for the bool flag kind of makes it seem like if it is false, then the function is a no-op, which isn't the case.

guziy:
Thanks for your comments @WeatherGod:


> I'd feel better about this if there were some unit tests added
I understand. I'll add few tests.

>  what was the reason for excluding scatter() from shiftdata?

scatter is not completely excluded from the shiftdata, but the reindexing (the reindexing which is supposed to set lower indices for the points with coordinates to the right of lon_0 + 180) is not performed in the case of scatter. This reindexing is important when a field is plotted and is based on the fact that the neighbouring points are close, which is not necessarily true for scatter. With the modification, the only thing done in case of scatter is converting longitudes to the (lon_0 -180, lon_0 + 180) range, so that it is still consistent with pcolormesh or contourf results..


>Also, and perhaps this is because the docstring for shiftdata isn't all that great, but an initial read of the doc for the bool flag kind of makes it seem like if it is false, then the function is a no-op, which isn't the case.

I'll try to make it clearer.

Cheers

WeatherGod:
Awesome work!

Issue #333, opened 2017-01-18
zorder failure with matplotlib 2.0.0 and current basemap master


akrherz:
Previously, I was able to set a `zorder=3` on things like `drawstates()` and then later run `contourf()` with a `zorder=2` and the plot would appear correctly.  For example:
```python
import matplotlib.pyplot as plt
from mpl_toolkits.basemap import Basemap
import numpy as np
from scipy.interpolate import NearestNDInterpolator

m = Basemap(llcrnrlon=-100, llcrnrlat=38,
            urcrnrlon=-86, urcrnrlat=46)

m.drawstates()

lons = np.arange(-110, -80, 1)
lats = np.arange(30, 60, 1)
vals = np.arange(30)
xi = np.linspace(-120, -60, 100)
yi = np.linspace(23, 50, 100)
xi, yi = np.meshgrid(xi, yi)
nn = NearestNDInterpolator((lons, lats), vals)
vals = nn(xi, yi)
m.drawstates(zorder=3)
m.contourf(xi, yi, vals, 20, zorder=2)
plt.savefig('test.png')
```
This code appears to work on matplotlib 1.5.3 and 2.0.0 with a master build from approximately 24 Oct 2016.  As I try current basemap master build with matplotlib 2.0.0, it fails.  Failing meaning that the contourf appears overtop the states. For example:

```
m.contourf(xi, yi, vals, 20, zorder=2)
m.drawstates(zorder=3)
```
![test](https://cloud.githubusercontent.com/assets/210858/22069375/d6637000-dd5d-11e6-9dc2-3bb6bba15b01.png)

```
m.drawstates(zorder=3)
m.contourf(xi, yi, vals, 20, zorder=2)
```
![bad](https://cloud.githubusercontent.com/assets/210858/22069381/da0aef6c-dd5d-11e6-8211-0d7257367db4.png)

Thank you!

akrherz:
I have isolated the problem to recent changes done with `ax._hold`.  I have no experience with `_hold`, but will look further into it.

WeatherGod:
attn: @efiring

akrherz:
submitted PR with what appears to fix my issue, but I am not fully certain if this works with matplotlib < 2, sorry.

WeatherGod:
closed by #334

PR #334, opened 2017-01-18
fix _hold assignment see #333

akrherz:


WeatherGod:
yup! looks like I didn't review the previous pull request closely enough. Good catch!

efiring:
Thanks for finding and fixing that.

guziy:
It's amazing, It is hard to believe that this would cause the #333 ))

Cheers

PR #337, opened 2017-02-02
Fix AttributeError: module 'matplotlib.colors' has no attribute 'normalize'

cgohlke:
Using matplotlib 2.0:
```
Running plotmap_masked.py
  File "D:\Build\basemap\basemap-git\examples\plotmap_masked.py", line 43, in <module>
    im = m.imshow(topodatm,palette,norm=colors.normalize(vmin=0.0,vmax=3000.0,clip=False))
AttributeError: module 'matplotlib.colors' has no attribute 'normalize'
TEST FAILURE (status=1)
```

See also http://matplotlib.org/api/colors_api.html#matplotlib.colors.Normalize

### DIFF
--- a/examples/plotmap_masked.py
+++ b/examples/plotmap_masked.py
@@ -40,7 +40,7 @@ topodatm = ma.masked_values(topodat, 1.e10)
 palette = plt.cm.YlOrRd
 palette.set_bad('aqua', 1.0)
 # plot image over map with imshow.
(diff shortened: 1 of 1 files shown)


WeatherGod:
how come we didn't notice this before?

WeatherGod:
Found it... it was deprecated in v1.3 of matplotlib and finally removed.

PR #339, opened 2017-02-02
Fix RuntimeWarning and DeprecationWarning in test.py

cgohlke:
Fixes
```
Running test.py
<snip>
D:\Build\basemap\basemap-git\examples\test.py:380: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
<snip>
X:\Python36\lib\site-packages\matplotlib\pyplot.py:524: RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`).
  max_open_warning, RuntimeWarning)
```

### DIFF
--- a/examples/test.py
+++ b/examples/test.py
@@ -5,6 +5,9 @@ from __future__ import print_function
 # country boundaries, filling continents and drawing
 # parallels/meridians
 
+import matplotlib
+matplotlib.rcParams['figure.max_open_warning'] = 0
+
 from mpl_toolkits.basemap import Basemap, cm, shiftgrid
 import numpy as np
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


examples/test.py
```diff
@@ -5,6 +5,9 @@
 # country boundaries, filling continents and drawing
 # parallels/meridians
 
+import matplotlib
+matplotlib.rcParams['figure.max_open_warning'] = 0
```
WeatherGod: This should have a test that the parameter exists first before trying to set it. This would error out for older versions of matplotlib.

PR #365, opened 2017-07-25
Add rather than nest include arrays for geos

Russell-Jones-OxPhys:
Shouldn't the contents of inc_dirs be added to geos_include_dirs rather than nested within it?

When I run "python3 setup.py build", building _geoslib.c fails as one of the -I flags is
`-I['/.../lib/python3.4/site-packages/numpy/core/include'] `

I got the same for `python3 -m pip install --upgrade -e 'git+https://github.com/matplotlib/basemap.git#egg=basemap'` but it seems an editable install is not possible because of how mpl_toolkits.axes_grid1 is imported.


WeatherGod:
Huh... curious that that ever even worked in the first place...

PR #366, opened 2017-08-07
Updated fillstates.py to add Alaska and Hawaii as map insets

jsh9:
I modified the example so that the map output imitates the map shown here (basically, Alaska and Hawaii are shown on the bottom left corner of the main map as small insets): https://en.wikipedia.org/wiki/List_of_U.S._states_by_population_density

I based my modifications partly on some code snippets in this stackoverflow thread (https://stackoverflow.com/questions/39742305/how-to-use-basemap-python-to-plot-us-with-50-states).

### DIFF
--- a/examples/fillstates.py
+++ b/examples/fillstates.py
@@ -6,6 +6,8 @@ from matplotlib.colors import rgb2hex, Normalize
 from matplotlib.patches import Polygon
 from matplotlib.colorbar import ColorbarBase
 
+fig, ax = plt.subplots()
+
 # Lambert Conformal map of lower 48 states.
 m = Basemap(llcrnrlon=-119,llcrnrlat=20,urcrnrlon=-64,urcrnrlat=49,
             projection='lcc',lat_1=33,lat_2=45,lon_0=-95)
@@ -93,9 +95,6 @@ for shapedict in m.states_info:
     statenames.append(statename)
 
 #%% ---------  cycle through state names, color each one.  --------------------
-fig = plt.gcf()  # get current figure instance
-ax = plt.gca() # get current axes instance
-
 for nshape,seg in enumerate(m.states):
     # skip DC and Puerto Rico.
     if statenames[nshape] not in ['Puerto Rico', 'District of Columbia']:
@@ -105,16 +104,21 @@ for nshape,seg in enumerate(m.states):
 
 AREA_1 = 0.005  # exclude small Hawaiian islands that are smaller than AREA_1
 AREA_2 = AREA_1 * 30.0  # exclude Alaskan islands that are smaller than AREA_2
-AK_SCALE = 0.18  # scale down Alaska to show as a map inset
+AK_SCALE = 0.19  # scale down Alaska to show as a map inset
+HI_OFFSET_X = -1900000  # X coordinate offset amount to move Hawaii "beneath" Texas
+HI_OFFSET_Y = 250000    # similar to above: Y offset for Hawaii
+AK_OFFSET_X = -250000   # X offset for Alaska (These four values are obtained
+AK_OFFSET_Y = -750000   # via manual trial and error, thus changing them is not recommended.)
 
 for nshape, shapedict in enumerate(m_.states_info):  # plot Alaska and Hawaii as map insets
     if shapedict['NAME'] in ['Alaska', 'Hawaii']:
         seg = m_.states[int(shapedict['SHAPENUM'] - 1)]
         if shapedict['NAME'] == 'Hawaii' and float(shapedict['AREA']) > AREA_1:
-            seg = list(map(lambda (x,y): (x-1900000, y+250000), seg))
+            seg = [(x + HI_OFFSET_X, y + HI_OFFSET_Y) for x, y in seg]
             color = rgb2hex(colors[statenames[nshape]])
         elif shapedict['NAME'] == 'Alaska' and float(shapedict['AREA']) > AREA_2:
-            seg = list(map(lambda (x,y): (AK_SCALE*x-200000, AK_SCALE*y-650000), seg))
+            seg = [(x*AK_SCALE + AK_OFFSET_X, y*AK_SCALE + AK_OFFSET_Y)\
+                   for x, y in seg]
             color = rgb2hex(colors[statenames[nshape]])
         poly = Polygon(seg, facecolor=color, edgecolor='gray', linewidth=.45)
         ax.add_patch(poly)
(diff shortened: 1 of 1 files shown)


story645:
I love this in theory 'cause I forgot too late I left out states, but is there anyway to refine this example so it's not quite this complicated?

jsh9:
I thought about it and I don't think the code can be simplified much while maintaining the look of the map.

I added three elements:
(1) Alaska and Hawaii
(2) Light gray lines to bound Alaska and Hawaii so that people know that they are map insets
(3) Color bar showing what each color means

Each of the three elements are essential to a scientifically complete map, and the codes I wrote for each of the three elements are already quite simple.  Also, for (1) and (2), I choose to use Mercator projection, because (a) this projection makes Alaska look "upright" instead of "tilted", and (b) enables me to draw straight gray lines on the map.

examples/fillstates.py
```diff
@@ -80,18 +89,59 @@
...
+m_.plot(np.linspace(180,180),np.linspace(26,23),linewidth=1.,
+        color=light_gray,latlon=True)
+m_.plot(np.linspace(-180,-177),np.linspace(23,20),linewidth=1.,
+        color=light_gray,latlon=True)
+m_.plot(np.linspace(-180,-175),np.linspace(26,26),linewidth=1.,
+        color=light_gray,latlon=True)
+m_.plot(np.linspace(-175,-171),np.linspace(26,22),linewidth=1.,
+        color=light_gray,latlon=True)
+m_.plot(np.linspace(-171,-171),np.linspace(22,20),linewidth=1.,
+        color=light_gray,latlon=True)
```
WeatherGod: These plot commands are distracting and largely overkill anyway. The linspace() doesn't really make sense since all you need are the vertexes. This could be simplified into two plot() calls, or possibly even two Polygons with `'none'` as their facecolor?

examples/fillstates.py
```diff
@@ -80,18 +89,59 @@
...
+
+#%% ---------   Show color bar  ---------------------------------------
+ax_c = fig.add_axes([0.9, 0.1, 0.03, 0.8])
+cb = ColorbarBase(ax_c,cmap=cmap,norm=norm,orientation='vertical',
+                  label=r'[population per $\mathregular{km^2}$]')
```
WeatherGod: Why not a regular colorbar creation approach instead of manually creating axes and such? An example like this is already getting pretty complicated, so we should aim to reduce complexity where-ever possible.

examples/fillstates.py
```diff
@@ -80,18 +89,59 @@
...
+AK_SCALE = 0.18  # scale down Alaska to show as a map inset
+
+for nshape, shapedict in enumerate(m_.states_info):  # plot Alaska and Hawaii as map insets
+    if shapedict['NAME'] in ['Alaska', 'Hawaii']:
+        seg = m_.states[int(shapedict['SHAPENUM'] - 1)]
+        if shapedict['NAME'] == 'Hawaii' and float(shapedict['AREA']) > AREA_1:
+            seg = list(map(lambda (x,y): (x-1900000, y+250000), seg))
+            color = rgb2hex(colors[statenames[nshape]])
+        elif shapedict['NAME'] == 'Alaska' and float(shapedict['AREA']) > AREA_2:
+            seg = list(map(lambda (x,y): (AK_SCALE*x-200000, AK_SCALE*y-650000), seg))
```
WeatherGod: Why not a list comprehension? Might be easier to read.

Also, could you include a comment to explain how these numbers came about (so that others could adapt this process for other projections, perhaps)?

examples/fillstates.py
```diff
@@ -2,18 +2,27 @@
...
-# http://en.wikipedia.org/wiki/List_of_U.S._states_by_population_density
+
+# Mercator projection, for Alaska and Hawaii
+m_ = Basemap(llcrnrlon=-190,llcrnrlat=20,urcrnrlon=-143,urcrnrlat=46,
+            projection='merc',lat_ts=20)  # do not change these numbers
+
+#%% ---------   draw state boundaries  ----------------------------------------
+## data from U.S Census Bureau
+## http://www.census.gov/geo/www/cob/st2000.html
+shp_info = m.readshapefile('st99_d00','states',drawbounds=True,
```
story645: can this be read in only once and the drawbounds be a part of drawing the polygons?

examples/fillstates.py
```diff
@@ -80,18 +89,59 @@
         # calling colormap with value between 0 and 1 returns
         # rgba value.  Invert color range (hot colors are high
         # population), take sqrt root to spread out colors more.
-        colors[statename] = cmap(1.-np.sqrt((pop-vmin)/(vmax-vmin)))[:3]
+        colors[statename] = cmap(np.sqrt((pop-vmin)/(vmax-vmin)))[:3]
     statenames.append(statename)
-# cycle through state names, color each one.
+
+#%% ---------  cycle through state names, color each one.  --------------------
+fig = plt.gcf()  # get current figure instance
```
story645: `fig, ax = plt.subplots()` at the top and then pass the ax kwarg to the Basemap methods?

examples/fillstates.py
jsh9: Plotting lines on a map with Mercator projection ("m_" in the code) using m_.plot([x1,x2],[y1,y2]) leads to index error:
**IndexError: index -2 is out of bounds for axis 0 with size 1**
I do not know the reason for this error, so I had to use np.linspace().

examples/fillstates.py
jsh9: Not really. Alaska and Hawaii are plotted using Mercator projection, while the other contiguous states are plotted using Lambert projection. The reason for choosing Mercator projection to plot Alaska and Hawaii is purely an aesthetic one, and I have noticed that this map-making choice (i.e., Alaska and Hawaii being plotted by Mercator projection) was adopted from as early as half a century ago.

I am not sure what you meant by "drawbounds be a part of drawing the polygons". Line 20 is in the original example script.

examples/fillstates.py
jsh9: The original example script has `ax = plt.gca()` on that location, and I just added `fig = plt.gcf()` on top of it because I wanted the changed to the original script to be minimal.

I can change it to what you suggested. It is indeed more elegant.

examples/fillstates.py
jsh9: Yeah, I will change them to list comprehension. I copied and pasted the list(map(lambda())) directly from here https://stackoverflow.com/questions/39742305/how-to-use-basemap-python-to-plot-us-with-50-states without much thinking

These number are achieved by trial and error; I will add comments to point this out.

examples/fillstates.py
jsh9: I tried the regular `m.colorbar()` way, but I couldn't figure out what I should use for `mappable`. Please provide advice if you can. Thanks.

examples/fillstates.py
jsh9: I modified these lines into four lines.

However, there is a tiny difference between the new result and old result: the light gray lines in the new version are now slightly thicker and look more "blurred" than the old version...

jsh9:
Made a new commit just now. Suggestions welcome!

examples/fillstates.py
guziy: Hi @jsh9:

I have this file around where I was coloring countries based on population. I used it for teaching purposes, so there is an intentional bug somewhere, but you can check out how I did the colorbar there at the end:

https://github.com/guziy/CMOS-python-tutorial/blob/master/fiona_test.py

Cheers

WeatherGod:
Sorry this sat for so long! Hurricane season distracted me and I forgot to get back to this. Thanks for your contribution!

Issue #177, opened 2015-02-20
PIL v.s Pillow


HughP:
It seems that PIL, the graphics library dependency is dead (or near dead) and modern development is on [Pillow](https://github.com/python-pillow/Pillow). I am new to the Python landscape. So I [asked over at homebrew](https://github.com/Homebrew/homebrew-python/issues/204) if there might be a brew for PIL but evidently not. I am just wondering if there is a way to force basemap to use Pillow instead of PIL. Several of the functions fail saying they need PIL, but others work.

jenshnielsen:
As far as I know pillow should be mostly a drop in replacement for PIL. You can install pillow and every thing should be fine.

http://pillow.readthedocs.org/porting-pil-to-pillow.html 
The only changes needed should be to ensure that Image is imported from PIL ands that `_imaging` is not imported.  I can't find any occurrences of these in the Basemap code so I think the only thing left to do is to ensure that the documentation is up to data.

mdboom:
matplotlib itself has supported Pillow for some time.

WeatherGod:
"Several of the functions fail saying they need PIL, but others work."

It would help to know which ones fail, and what their tracebacks are.

HughP:
Which ones work (**or don't work**)? I have tried the robinson projection and plotting a dataset, [following this tutorial](http://introtopython.org/visualization_earthquakes.html)... I managed to produce a map with my data plotted, [my successful image is here](https://github.com/HughP/Lexical-Database-Archiving-Stats/blob/master/2015-Hawaii/Formulas/Language_Plot_RandC.png): ![Lexical Database Archiving](https://github.com/HughP/Lexical-Database-Archiving-Stats/blob/master/2015-Hawaii/Formulas/Language_Plot_RandC.png)

I wanted to add some texture to my image so I thought I might try `map.bluemarble()` or `map.shadedrelief()` and in both cases I get a message saying that PIL was not installed, which is of course true, because I have brewed and am using Pillow 2.7.0. As you can see, I have gotten some image out of basemap, so I assume I am doing something correctly. Then [I consulted the documentation](http://matplotlib.org/basemap/users/installing.html) and saw that it said:
>**Optional libraries**

>PIL
>Python Imaging Library (download), only needed for `bluemarble()`, `etopo()`, `shadedrelief()` and `warpimage()` instance methods.

I hit two of four and the error message and thought I would ask.

[...]

PR #388, opened 2018-01-19
mention pillow when PIL is not found (instead of linking to long-dead PIL page)

megies:
I think instead of advertising the long-stale PIL page the import error should directly point towards pillow.

see also #177


WeatherGod:
Thanks!

PR #402, opened 2018-04-27
Setuptools

ocefpaf:
~~# Do not merge this before #400 and #401~~

... and simplify how we deal with the requirements in `setup.py`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,8 +9,6 @@ from setuptools.dist import Distribution
 if sys.version_info < (2, 6):
     raise SystemExit("""matplotlib and the basemap toolkit require Python 2.6 or later.""")
 
-from setuptools.dist import Distribution
-
 # Do not require numpy for just querying the package
 # Taken from the netcdf-python setup file (which took it from h5py setup file).
 inc_dirs = []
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,20 +1,22 @@
...
+import sys
+from setuptools.dist import Distribution
 
 if sys.version_info < (2, 6):
     raise SystemExit("""matplotlib and the basemap toolkit require Python 2.6 or later.""")
 
-from distutils.dist import Distribution
-from distutils.util import convert_path
-from distutils import ccompiler, sysconfig
+from setuptools.dist import Distribution
```
efiring: Duplicated above.

setup.py
```diff
@@ -23,6 +25,21 @@
...
+        return fp.read()
+
+
+def get_install_requirements(path):
+    content = read(path)
+    return [
+        req
+        for req in content.split("\n")
+        if req != '' and not req.startswith('#')
+    ]
```
efiring: style quibble: I don't like the closing paren being indented that way.  I would probably put the `read` functionality inside the `get_install_requirements` function, since it isn't used anywhere else, and then format the return as
```python
    return [req for req in content.split("\n")
                if req and not req.startswith('#')]
```

WeatherGod:
could you rebase this onto master to make it easier to see the remaining differences?

ocefpaf:
This is ready for another round of review.

WeatherGod:
Is it basically assumed now that people have setuptools installed? I guess it is pretty much safe because matplotlib has been requiring setuptools for awhile now, right?

ocefpaf:
> Is it basically assumed now that people have setuptools installed?

Yep. `pip` is the recommended way to install packages now and `pip` brings `setuptools`. Also, `pip` is now bundle with all basic Python distributions, so it is always expected to be present.

Issue #405, opened 2018-05-14
new release?


sandrotosi:
GH report 48 commits since 1.1.0 was tagged, maybe it's time for a new release of basemap?

megies:
Totally +1 especially since 1.1.0 will break with matplotlib 2.3, see #382.

fjansson:
On this note, would it be possible to get a new release in Pypi also? The version there is still 1.0.7.
I just hit matplotlib issue https://github.com/matplotlib/basemap/issues/395 - basemap 1.0.7 apparently does not work with matplotlib 2.2.2, both of which I installed with pip.

I see this has been discussed elsewhere too, though: https://github.com/matplotlib/basemap/issues/251

hany:
Hi there,

Just ran into #395 as well. Wondering when a new version will be pushed to PyPI?

gwerbin:
Installing basemap 1.2 with Conda actually downgraded my Matplotlib all the way to 1.5. Yikes.

gwerbin:
No. I think my setup was just mildly "broken", I have a ton of packages installed and I'm deep in dependency hell anyway.

xylar:
With the merge of #454, it would be especially nice to have a release of basemap that supports the latest proj4/pyproj.  basemap is starting to cause major compatibility problems with other packages on conda-forge, which have moved on to pyproj>=2 and proj4>=6.

molinav:
There is a new release available (1.3.0). This release also brings some precompiled binary wheels to PyPI, and `basemap` is split in three parts in a similar way to what Anaconda does. With `pip`, now it should be as simple as:
```sh
python -m pip install basemap
```
and, in case that you also need the high-resolution datasets:
```sh
python -m pip install basemap-data-hires
```

Issue #403, opened 2018-04-30
publish v1.1.0 to pypi


alaniwi:
1.1.0 has been tagged in github but isn't being picked up by default during pip install (latest on pypi seems to be 1.0.7). Unless there is a good reason not to, please can it be published to pypi?

t-makaro:
And as a conda package too. Currently, the only conda packages of basemap 1.0.7 are for python 2.7 and Numpy1.10, but I'd like to be able to install on python 3.6

WeatherGod:
The "official" conda packages for basemap is on the conda-forge channel. I
have no control over the packages on the standard channel. The conda-forge
channel pretty much tracks the master branch (not quite, but it gets all of
the important stuff).

heitorPB:
Could you publish version 1.2.0 on pypi?

molinav:
@heitorPB If you are using GNU/Linux, you may find pre-compiled binary wheels in PyPI for the current beta release (1.3.0b1), I just uploaded them yesterday. The stable release 1.3.0 will come with some reorganisation (splitting `basemap` into `basemap`, `basemap-data` and `basemap-data-hires`). For the moment you could do:
```sh
python -m pip install "basemap-data == 1.3.0b1" "basemap == 1.3.0b1"
```
You only need `basemap-data-hires` if you need the high-resolution data assets (i.e. those labelled with "h" and "f").

molinav:
@heitorPB I just bring you the nice update that we have stable release for `basemap` version 1.3.0. So now we can just install `basemap` with `pip` as we would do with any other package:
```sh
python -m pip install basemap
```

heitorPB:
Thanks!

Issue #251, opened 2016-01-05
basemap is not found on pip for python3.4?


guziy:
Hi:

I've recently tried to install basemap using pip on my fedora 23 in py3.4 virtual environment, but:

```!bash
(py3.4) huziy at thinkpad in ~/Python/Projects/PyNotebooks on master*
$ pip install basemap
Collecting basemap
  Could not find a version that satisfies the requirement basemap (from versions: )
No matching distribution found for basemap
(py3.4) huziy at thinkpad in ~/Python/Projects/PyNotebooks on master*
$ python
Python 3.4.3 (default, Jun 29 2015, 12:16:01) 
[GCC 5.1.1 20150618 (Red Hat 5.1.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
```
Which is weird, since it worked OK before...

Maybe the versions should be specified in the setup.py?

Cheers

tacaswell:
What do you mean by 'before'?  Did it used to work with 3.4 and now it does not?

guziy:
Yes: I've used it with python3.4 before ...

2016-01-04 23:36 GMT-05:00 Thomas A Caswell <notifications@github.com>:

> What do you mean by 'before'? Did it used to work with 3.4 and now it does
> not?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/matplotlib/basemap/issues/251#issuecomment-168894629>.
>



-- 
Sasha

tacaswell:
Interesting.  I am also seeing the same issue.

jenshnielsen:
This is a duplicate of #198 Basemap is to large to be hosted on Pypi so the package is hosted externally and new versions of pip no longer allows that pr default.

guziy:
@jenshnielsen:
so this stopped working for the recent pip?
```!bash
$ pip3 install --allow-external basemap --allow-unverified basemap basemap
Collecting basemap
  Could not find a version that satisfies the requirement basemap (from versions: )
No matching distribution found for basemap

```

jenshnielsen:
Im not sure someone may also have removed the links to sourceforge from Pypi

aeroevan:
It looks like allow external was removed from recent versions of pip:
http://legacy.python.org/dev/peps/pep-0470/

Something like pip install https://github.com/matplotlib/basemap/archive/v1.0.7rel.tar.gz should work though.

robertjliguori:
Yes, I can't get PyCharm to install it either. :(

[...]

Issue #198, opened 2015-08-09
Consider moving the pip package away from sourceforge


FLamparski:
Currently, the url for basemap in pypi points at <https://downloads.sourceforge.net/project/matplotlib/matplotlib-toolkits/basemap-1.0.7/basemap-1.0.7.tar.gz>. This means that in order to install basemap, one has to type the rather scary-looking and lengthy `pip3 install --allow-external basemap --allow-unverified basemap basemap`. Given this and some relatively shady stuff sourceforge have engaged in recently, I think it's a good idea for you to move the distribution and update the pypi listing accordingly.

markmuetz:
I've found this too, it makes installation of basemap a bit of a pest when using pip. And I'd echo the sentiments about sourceforge as well.

WeatherGod:
These are valid points, and it is something to consider. Previously, the
limitation in hosting the basemap package on pypi has been its size.
Perhaps that can be revisited?

On Tue, Aug 11, 2015 at 5:17 AM, Mark Muetzelfeldt <notifications@github.com

matthew-brett:
Relevant thread : https://github.com/matplotlib/matplotlib/issues/2715  Maybe file size limit of 50MB?

WeatherGod:
Thread rezz-ing.
@dstufft, you have helped us out before with file-size limits for matplotlib on PyPi (https://github.com/matplotlib/matplotlib/issues/2715). The problem is significantly worse for basemap, as the tarball is just under 130MB. What are our options?

dstufft:
If you can wait until Warehouse is ready (scheduled to be early 2016), we'll have the ability to allow larger file sizes for specific projects so we can whitelist particular ones to have larger upload limits (our theoretical maximum will be several GB... but at some point we start running into problems with the cost of storing multiple releases of giant packages). In any case, 130MB for a well established project like matplotlib shouldn't be an issue once we can loosen the restrictions on a per project basis.

fbender:
@dstufft Is Warehouse ready to be used for basemap now? As far as I can see it's still in development but at least partially operational.
@WeatherGod Would it be a good idea to test hosting basemap there now?

dstufft:
Yea, you can upload to Warehouse now, but you'll want to be aware of pypa/warehouse#1194.

WeatherGod:
Is that true even if the package was never hosted on pypi directly? basemap
was always hosted from sourceforge.

dstufft:
[...]

PR #433, opened 2018-10-21
Add installation instructions for pip

heitorPB:
Several issues report problems with using `pip install basemap`. A workaround is using the latest version from Github.

Some issues with pip problems: #405 #403 #198 #251

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,15 @@ prompt.
 To run all the examples (except those that have extra dependencies
 or require an internet connection), execute 'python run_all.py'.
 
+An alternative method is using `pip`:
+
+```
+pip install git+https://github.com/matplotlib/basemap.git
+```
+
+You migh want to run pip with administrative privilegies (e.g. `sudo pip`) or
+perform an user only installation (e.g. `pip install --user`).
+
 ## Contact
 
 Ben Root <ben.v.root@gmail.com>
--- a/README.md
+++ b/README.md
@@ -3,6 +3,9 @@
 Plot on map projections (with coastlines and political boundaries)
 using matplotlib.
 
+**Warning**: this package is being deprecated in favour of
+[cartopy](https://scitools.org.uk/cartopy/docs/latest/).
+
 ## Requirements
 
 * Python 2.6 (or higher)
--- a/README.md
+++ b/README.md
@@ -103,11 +103,8 @@ or require an internet connection), execute 'python run_all.py'.
 An alternative method is using `pip`:
 
 ```
-pip install git+https://github.com/matplotlib/basemap.git
+pip install --user git+https://github.com/matplotlib/basemap.git
 ```
-
-You migh want to run pip with administrative privilegies (e.g. `sudo pip`) or
-perform an user only installation (e.g. `pip install --user`).
 
 ## Contact
 


README.md
```diff
@@ -100,6 +100,15 @@ prompt.
 To run all the examples (except those that have extra dependencies
 or require an internet connection), execute 'python run_all.py'.
 
+An alternative method is using `pip`:
+
+```
+pip install git+https://github.com/matplotlib/basemap.git
+```
+
+You migh want to run pip with administrative privilegies (e.g. `sudo pip`) or
```
phobson: I don't think we should encourage `sudo pip`. Let's only promote the `--user` option.

heitorPB:
Updated with @phobson suggestion.

WeatherGod:
Does this method work in a clean environment? How does it handle the libgeos stuff?

To be honest, I'd rather promote the conda-forge approach to installing basemap. Or, even better, `pip install cartopy`.

heitorPB:
Assuming you have pyproj installed, the pip command works. pyproj also has a `pip install` problem. I understood this method compiles the libgeos stuff included in the repo.

What is the advantage of conda-forge? IMHO, having several package managers is a nightmare. A quick browser in coda's documentation showed me that most packages I use are not available there. And they are in pip.

ocefpaf:
> Assuming you have pyproj installed, the pip command works. pyproj also has a `pip install` problem. I understood this method compiles the libgeos stuff included in the repo.

Ideally we should start building wheels for `pyproj` to facilitate the life of `pip` users. Basemap I'm not so sure due to the deprecation. We don't want to encourage people to use something that is at the end life support.

> What is the advantage of conda-forge?

It is a collective of packages (5508) managed openly by the community:

https://anaconda.org/conda-forge/repo

> IMHO, having several package managers is a nightmare.

Indeed. That is why people are moving to `conda` and using `conda-forge` for packages. It is platform and language agnostic. As it evolves it becomes the only package manager you will ever need. (Still rough around the edges with haskell, rust, and go. But great for Python and R.)

> A quick browser in coda's documentation showed me that most packages I use are not available there. And they are in pip.

Not sure what you are using but take a look at the `conda-forge` channel package listing and not `conda`'s documentation. Also, if a package is not there. It is relatively easy to add.

heitorPB:
Oh, Basemap is being deprecated?



> Ideally we should start building wheels for `pyproj` to facilitate the life of `pip` users.

I submitted a PR there with instruction to `pip install` pyproj from github as well. But wheels would be better.

ocefpaf:
> Oh, Basemap is being deprecated?

See the @WeatherGod's (the basemap maintainer) https://github.com/matplotlib/basemap/pull/433#issuecomment-432298598 and the EOL docs: https://matplotlib.org/basemap/users/intro.html#cartopy-new-management-and-eol-announcement

> I submitted a PR there with instruction to `pip install` pyproj from github as well. But wheels would be better.

pip install from github and expecting to have the libs and compilers is too advanced for the average user. Building wheels are not that hard with the new technologies, but unfortunately that is up to the PyPI package authors/maintainers and many do not have the time to do that. That is why `conda-forge` is better, it is a community that not makes packaging easy to anyone who wishes to crate or maintain, not only the package authors/maintainers. Also, b/c most of the libs, like `geos`, `gdal`, `proj.4`, etc, are already package for Linux, Windows and OS X, so creating a python package that depended on those is not a burden.

matthew-brett:
I can't say whether conda-forge is better, but there are a lot of packages using (my) https://github.com/matthew-brett/multibuild framework for wheel building.  It's not completely trivial, but it's not that hard either.  I'd guess about 3 -4 hours for the initial setup.

README.md
denisarnaud: Install from the Git repositories work on MacOS Mojave with Python 3.7 and pipenv and/or pip.
Reference: https://stackoverflow.com/questions/51963619/pyproj-fails-to-compile-when-i-pip-install-it-and-its-not-about-gcc
* With ``pipenv``
```bash
$ pipenv install cython
$ pipenv install git+https://github.com/jswhit/pyproj.git#egg=pyproj
$ pipenv install git+https://github.com/matplotlib/basemap.git
```

* With just ``pip``:
```bash
$ pip install -U cython
$ pip install -U git+https://github.com/jswhit/pyproj.git#egg=pyproj
$ pip install -U git+https://github.com/matplotlib/basemap.git
```

ver0z:
I solved this by uninstalling the conda and reinstalling, defining on installation the conda path on variable environment. After, at settings on your project, set the project interpreter to C:\Users\the_user_name\anaconda3\python.exe, then try to install the basemap. It was the only way I got this, after many tries and bugs with pip I tried with conda and that's it.

PR #454, opened 2019-03-13
Fixed compatibility issue with pyproj 2.0+

pandrey-fr:
Starting with version 2.0, the `pyproj.pyproj_datadir` constant was deprecated in favor of the `pyproj.datadir.get_data_dir()` constant.
A minor alteration to Basemap's `__init__.py` file allows compatibility with both the newer and older versions of pyproj regarding this breaking change.

### DIFF
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -149,10 +149,27 @@ projection_params = {'cyl'      : 'corners only (no width/height)',
              }
 
 # create dictionary that maps epsg codes to Basemap kwargs.
-if hasattr(pyproj, 'datadir'):
-    epsgf = open(os.path.join(pyproj.datadir.get_data_dir(), 'epsg'))
-else:
-    epsgf = open(os.path.join(pyproj.pyproj_datadir, 'epsg'))
+if hasattr(pyproj, 'datadir'):  # pyproj version 2+
+    epsgf = os.path.join(pyproj.datadir.get_data_dir(), 'epsg')
+    if not os.path.isfile(epsgf):
+        if os.name == 'posix':
+            epsgf = os.environ.get('PROJ_LIB', '/usr/share/proj')
+        elif 'PROJ_LIB' in os.environ:
+            epsgf = os.environ['PROJ_LIB']
+        else:
+            raise FileNotFoundError(
+                ("Cannot find '%s'" % epsgf)
+                + "and PROJ_LIB environment variable is not set."
+            )
+        epsgf = os.path.join(epsgf, 'epsg')
+        if not os.path.isfile(epsgf):
+            raise FileNotFoundError(
+                ("Cannot find '%s'" % epsgf)
+                + "nor epsg file under pyproj's internal datadir."
+            )
+else:  # older versions of pyproj
+    epsgf = os.path.join(pyproj.pyproj_datadir, 'epsg')
+epsgf = open(epsgf)
 epsg_dict={}
 for line in epsgf:
     if line.startswith("#"):
--- a/lib/mpl_toolkits/basemap/__init__.py
+++ b/lib/mpl_toolkits/basemap/__init__.py
@@ -152,20 +152,23 @@ projection_params = {'cyl'      : 'corners only (no width/height)',
 if hasattr(pyproj, 'datadir'):  # pyproj version 2+
     epsgf = os.path.join(pyproj.datadir.get_data_dir(), 'epsg')
     if not os.path.isfile(epsgf):
+        raise_exception = (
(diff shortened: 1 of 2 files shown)


WeatherGod:
Hmm, doesn't seem to be working for some of the travis builds (other travis builds are failing due to not having Cython installed.... that seems to be unrelated, I hope).

pandrey-fr:
The Python 2.6 and 3.4 travis builds failed to install pyproj, which seems unrelated to my fix (although it is indeed an issue).

For the other travis builds, I think it comes from the fact that espg is not comprised in pyproj by default (I had this issue on my own machine, and copied the espg file from /usr/share/proj to the /share/proj subfolder of my pyproj installation, which on the other hand contained most of the other projection files).

So, basically, we have to look into what other things changed with pyproj and how to remediate the issue...

pandrey-fr:
An alternative could be to try and bypass the pyproj data directory in favor of either /usr/local/share/proj (on Linux) or the PROJ_LIB environment variable (on other platforms or in case the former does not exist). This is in fact what proj.4 does, and what pyproj would do if it did not comprise its own data directory (see [this file](https://github.com/jswhit/pyproj/blob/master/pyproj/datadir.py)), which for some reason seems not to include espg, at least for the latest versions on Linux platforms.

pandrey-fr:
I updated the code with the following ideas, but the travis builds are still failing, as it appears that the epsg file is not found in any of the expected locations on the machines... I would say it is about a change in the installation of pyproj, but I have not looked it up yet.

cgohlke:
I don't think the `epsg` file is distributed with "standard" pyproj 2. Why not simply include the file in the basemap data directory?

pandrey-fr:
> I don't think the `epsg` file is distributed with "standard" PROJ 6 or pyproj 2. Why not simply include the file in the basemap data directory?

Thank you for confirming this; I will check the licensing of the file and commit a change to include it asap -- it the build passes, @WeatherGod will be able to judge whether it is pertinent to actually make this change to basemap or not.

pandrey-fr:
I have not had time to include the file (and change the code) yet, nor update legal mentions accordingly; but the terms of use of the espg seem to allow its distribution as part of basemap. They can be found [here](http://www.epsg.org/TermsOfUse.aspx).

pandrey-fr:
As expected, adding the file directly as part of basemap fixes the issue, and travis builds succeed - or fail for unrelated issues that seem to be caused by Cython when compiling either pyproj or (for the nightly python version) numpy. I am no expert in Travis, but I would guess it has to do with the virtual machines' configuration...

On the other hand, I did not include a proper licensing of the epsg file (and used one that may be out of date as I did not download the latest from the epsg website). @WeatherGod (or somebody else), would you take care of that, as I am no legal expert nor am entirely clear about the organization of legal mentions in the package?

pandrey-fr:
I went through the travis file and the builds' output stacks, and it appears that:
- 2.6 fails because of a Cython error
- nightly fails for similar reasons (maybe due to the installation of cython in "no-compile" mode?)
- 3.4 fails because it lacks a PROJ_DIR environment variable after starting to install pyproj, (which may be fixed by updating the mahcine's setup, adding an instruction to export said environment variable).

By the way, 2.6 is long-deprecated (as is 2.7, but I can get the point of maintaining it), and 3.4 has been deprecated after the update from March 18th, so do we really need to test for those systems, knowing that in fact people who might chose to stay on those systems can either stop updating their packages or figure out issues that are related to dependencies and not basemap itself?

WeatherGod:
So, the failures appear to be related to whenever we build using the internal libgeos. This is a completely separate problem, and I won't hold up this PR over it.

WeatherGod:
Where did you download the epsg file from? I need to determine its source in order to place the right license file in the source tree.

pandrey-fr:
> Where did you download the epsg file from? I need to determine its source in order to place the right license file in the source tree.

To be honest, I had it on my computer and do not remember exactly where it came from, but probably from an older version of PROJ.4. I went through the history of their repository, and found that the latest version to include epsg coordinates as a flat file is 5.2 ; I am therefore going to make a commit replacing the file I initially updated with the one found [here](https://github.com/OSGeo/proj.4/blob/5.2/nad/epsg).

On the other hand, I registered on the epsg website in hope to download the latest version of the file, but it now seems to only be distributed as a microsoft database file, which would need conversion to proj4 format - and I don't know how to do that.

pandrey-fr:
I think we should thus refer to PROJ.4's licensing, which can be found [here](https://github.com/OSGeo/proj.4/blob/5.2/COPYING) for the version I took the file from, and states that everything (including data files) can be re-used.

WeatherGod:
License file was added in PR #460

PR #473, opened 2019-08-20
port doc/make.py to py3k

sandrotosi:
we applied this patch in the last update to Debian, since the we are trying to remove python 2

### DIFF
--- a/doc/make.py
+++ b/doc/make.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from __future__ import print_function
 import fileinput
 import glob
 import os
--- a/doc/make.py
+++ b/doc/make.py
@@ -25,7 +25,7 @@ def latex():
 
         os.chdir('../..')
     else:
-        print 'latex build has not been tested on windows'
+        print('latex build has not been tested on windows')
 
 def clean():
     shutil.rmtree('build')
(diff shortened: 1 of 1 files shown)


doc/make.py
```diff
@@ -25,7 +25,7 @@ def latex():
 
         os.chdir('../..')
     else:
-        print 'latex build has not been tested on windows'
+        print('latex build has not been tested on windows')
```
WeatherGod: Need to add a `from __future__ import print_function` at the top so that this will still work in the python 2 series.

WeatherGod:
Keep in mind that there is no plan to drop python 2 and upgrade to py3k only because support for basemap will end in Jan. 2020.

sandrotosi:
I wasnt suggest to drop py2 at all, but you can build the documentation on py3k while still having the actual module support 2 and 3

WeatherGod:
There is no compelling reason to drop py2 doc building if the module itself is still written in python 2. Because there is no plan for moving on to py3k for basemap, dropping py2 doc-building seems entirely unnecessary.

WeatherGod:
Note, in case it wasn't clear, I am all for getting doc-building working on py3k. It is trivial to support both py2 and py3 for doc-building in this case.

doc/make.py
sandrotosi: done

WeatherGod:
Thanks for the patch. Note to self, I need to update the travis-ci file to unbreak the testing

PR #479, opened 2019-12-18
Update comment to match example code

obi1kenobi:
I'm not very familiar with this library, but this seemed to be a typo.


WeatherGod:
Indeed. Thank you.

PR #482, opened 2020-01-01
Add waning sign in README to emphasize deprecation

timhoffm:


timhoffm:
Self-merging as a trivial change.

Issue #495, opened 2020-08-02
basemap cannot be imported with matplotlib-3.3.0


joernu76:
To replicate, install fresh packages in conda

```
conda create -n basemap-test -c conda-forge basemap matplotlib
```

all packages:

```
 (e) basemap-test  ~  conda list                                     
# packages in environment at /home/icg173/anaconda3/envs/basemap-test:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                      1_llvm    conda-forge
basemap                   1.2.1            py37hd1be537_2    conda-forge
blas                      2.17                   openblas    conda-forge
ca-certificates           2020.6.20            hecda079_0    conda-forge
certifi                   2020.6.20        py37hc8dfbb8_0    conda-forge
cycler                    0.10.0                     py_2    conda-forge
dbus                      1.13.6               he372182_0    conda-forge
expat                     2.2.9                he1b5a44_2    conda-forge
fontconfig                2.13.1            h86ecdb6_1001    conda-forge
freetype                  2.10.2               he06d7ca_0    conda-forge
geos                      3.8.0                he1b5a44_1    conda-forge
gettext                   0.19.8.1          hc5be6a0_1002    conda-forge
glib                      2.65.0               h6f030ca_0    conda-forge
gst-plugins-base          1.14.5               h0935bb2_2    conda-forge
gstreamer                 1.14.5               h36ae1b5_2    conda-forge
icu                       64.2                 he1b5a44_1    conda-forge
intel-openmp              2019.4                      243  
jpeg                      9d                   h516909a_0    conda-forge
kiwisolver                1.2.0            py37h99015e2_0    conda-forge
lcms2                     2.11                 hbd6801e_0    conda-forge
ld_impl_linux-64          2.34                 hc38a660_9    conda-forge
libblas                   3.8.0               17_openblas    conda-forge
libcblas                  3.8.0               17_openblas    conda-forge
libclang                  9.0.1           default_hde54327_0    conda-forge
libedit                   3.1.20191231         h46ee950_1    conda-forge
libffi                    3.2.1             he1b5a44_1007    conda-forge
libgcc-ng                 9.3.0               h24d8f2e_13    conda-forge
libgfortran-ng            7.5.0               hdf63c60_13    conda-forge
libiconv                  1.15              h516909a_1006    conda-forge
liblapack                 3.8.0               17_openblas    conda-forge
[...]

PR #498, opened 2020-08-03
Update geos version number

joernu76:
Fix issue #496

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
 addons:
   apt:
     packages:
-      - libgeos-3.4.2
+      - libgeos-3.5.0
       # matplotlib requirements
       - libpng12-dev
       - libfreetype6-dev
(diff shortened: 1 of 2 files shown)


joernu76:
I removed 2.6, but kept 2.7 for now. I also added python 3.7+3.8 but removed 3.4 and 3.5.

Now the tests would likely pass when issue #495 would be adressed (all failing test cases show the "dedent" failure.

When accepting this and the #495 fix, travis will likely suceed, otherwise I can have another look.

WeatherGod:
Going to power-cycle the PR to see how it holds up now that I accepted #496.

WeatherGod:
Looks like something isn't right with libgeos v3.6.1. We are getting a segfault during the tests.

joernu76:
Why did the tests not segfault in the pipeline for my merge request?

joernu76:
Ah, because the dedent was fixed.
Will see what I can do to fix this.

joernu76:
The tests pass now. There was a "closed" issue pointing out this exact error. I modified the solution slightly. GEOS doesn't accept infinites any more and I added a workaround for this that has likely no implications. I do not fully understand the code at that point in time though. The fix cannot be worse than the crash, though...

WeatherGod:
And you see why I want to dump this codebase like the hot mess that it is?

Yeah, this fix looks reasonable, but it makes me worry that there are similar issues elsewhere in the codebase. The unit tests aren't nearly extensive enough. I am going to put it though some more paces by having it generate the documentation on my machine as well.

WeatherGod:
The examples are reminding me why I have been wanting to drop this library for so long, as some of them are broken for various reasons, but most are fine, and nothing seems broken by this fix, so it should be a net improvement. Merging.

PR #507, opened 2020-11-05
Set version number to 1.2.2

### LINKED ISSUE #506: Library version number for final release is wrong
The final release (1.2.2) keeps the previous version number (1.2.1) in the `__version__` variable inside the main file, so the users will download 1.2.2 but later will get:

```python
>>> from mpl_toolkits import basemap
>>> basemap.__version__
1.2.1
```

molinav:
This pull request fixes the `basemap` version number, which was not updated when releasing version 1.2.2.

Fixes #506.


molinav:
I close this pull request because my `master` branch has been advancing and now it would merge additional changes that are not related to the original version change.

Issue #512, opened 2021-03-02
ValueError: 'vertices' must be a 2D list or array with shape Nx2


flaviodoliveira:
I'm using Python 3.8.5
Matplotlib version 3.1.2
and basemap-1.2.2+dev

I'm running something like:
```

    llon = lons.min()
    llat = lats.min()
    ulon = lons.max()
    ulat = lats.max()

    fig, ax = plt.subplots()
    
    m = Basemap(llcrnrlon=llon, llcrnrlat=llat, urcrnrlon=ulon,
                urcrnrlat=ulat, resolution='h')

    m.drawcoastlines(linewidth=0.8)
    m.drawcountries(linewidth=0.8)
    m.drawstates(linewidth=0.8)


```
and then I got this:

>   File "plt_RAIN.py", line 62, in <module>
>     m.drawcoastlines(linewidth=0.8)
>   File "/usr/local/lib/python3.8/dist-packages/mpl_toolkits/basemap/__init__.py", line 1859, in drawcoastlines
>     coastlines = LineCollection(self.coastsegs,antialiaseds=(antialiased,))
>   File "/usr/lib/python3/dist-packages/matplotlib/collections.py", line 1333, in __init__
>     self.set_segments(segments)
>   File "/usr/lib/python3/dist-packages/matplotlib/collections.py", line 1348, in set_segments
>     self._paths = [mpath.Path(_seg) for _seg in _segments]
>   File "/usr/lib/python3/dist-packages/matplotlib/collections.py", line 1348, in <listcomp>
>     self._paths = [mpath.Path(_seg) for _seg in _segments]
>   File "/usr/lib/python3/dist-packages/matplotlib/path.py", line 129, in __init__
>     raise ValueError(
> ValueError: 'vertices' must be a 2D list or array with shape Nx2
> 

I checked everything, but I'm used to run this on basemap 1.0.7 and everything was ok.
And when I run without m.drawcoastlines(), there is no problem.

What can I do to run it?

guziy:
I have a similar problem. I think this might be because your domain does not intersect any coastlines? I have this reproducer:

```python
# fails 
from mpl_toolkits.basemap import Basemap
b = Basemap(projection="cea", llcrnrlon=267.5, urcrnrlon=283.995, llcrnrlat=41.35, urcrnrlat=51, resolution="l") 
b.drawcoastlines()
```

*Note*: The above fails in basemap 1.2.1 and works in basemap 1.1.0, regression somewhere...

[...]

PR #538, opened 2022-04-01
Fix drawcoastlines

guziy:
Fix https://github.com/matplotlib/basemap/issues/512


guziy:
Sorry I started from master and then rebased on develop, I think that introduced your commit in the pull request, do you want me to redo it ?

molinav:
@guziy No worries regarding the starting branch, it is not that important in this case. Thanks for the fix!

Issue #552, opened 2022-08-31
readshapefile: add an argument to override unicode decoding error handling


guziy:
Usually, we want to plot the shapes and not care too much about the correctness of decoded .dbf file.  But the current behavior: plotting fails if the default encoding does not match the one used for saving .dbf file.

I propose to add an argument to the readshapefile:

```python
mp.readshapefile(..., encoding_errors="strict") # I guess the possible values are  'ignore', 'replace' and 'backslashreplace'
```

Currently, this is hardcoded to 'strict' inside the `readshapefile` method (i.e. default in the `Reader` constructor).

I could try to submit a PR at some point.

PR #554, opened 2022-09-01
Add an option encoding_errors to readshapefile

guziy:
* to be able to ignore encoding errors
* default is strict

addresses #552


molinav (approved):

molinav:
It should be harmless to merge this feature, since it adds a new argument and defaults to the former behaviour, right?

So far it looks good to me. Being a "new feature", I would then bump the `basemap` version to 1.4.0 soon. But before that I had in mind to fix some of the remaining bugs in the 1.3.x series, just that I didn't find a free spot in my calendar to do it.

guziy:
Thanks @molinav: Yes to your question, this simply adds a new argument with the default value `strict` which is the current default in pyshp. To avoid the trouble to search for the right encoding for those who don't need the actual strings from the .dbf, they could just set `encoding_errors="ignore"`.

molinav:
Thanks for the contribution and the feedback, @guziy!

PR #562, opened 2022-10-30
Add caching to arcgisimage

JoepdeJong:
arcgisimageis slow due to network communication, especially when you are working with high-resolution plots of the same region. This commit adds a cachedir to the arcgisimage function, to prevent downloading the same image multiple times.

### DIFF
--- a/packages/basemap/src/mpl_toolkits/basemap/__init__.py
+++ b/packages/basemap/src/mpl_toolkits/basemap/__init__.py
@@ -4217,7 +4217,7 @@ class Basemap(object):
 
     def arcgisimage(self,server='http://server.arcgisonline.com/ArcGIS',\
                  service='World_Imagery',xpixels=400,ypixels=None,\
-                 dpi=96,verbose=False,**kwargs):
+                 dpi=96,cachedir=None,verbose=False,**kwargs):
         """
         Retrieve an image using the ArcGIS Server REST API and display it on
         the map. In order to use this method, the Basemap instance must be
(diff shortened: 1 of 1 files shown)


molinav:
Thanks for the contribution, @JoepdeJong! Since this pull request adds a new feature, I moved the target branch from `master` to `develop`, so that this new argument in `Basemap.arcgisimage` will appear in the next minor release of `basemap` (1.4.0).

Currently, I am putting most of my effort in the `basemap` 1.3.x series, with patches that were required since long time ago to make `basemap` a bit more stable, so I still do not have a planned date to prepare the release 1.4.0. In the meantime, you can install `basemap` directly from the `develop` branch with:
```
python -m pip install git+https://github.com/matplotlib/basemap#subdirectory=packages/basemap
```

Note that in this case you will need to have a C compiler for the `_geoslib` module build, and define `GEOS_DIR` in case that you do not have GEOS in some of the common installation paths.

molinav (approved):

guidocioni:
Hey @molinav, as I see this PR is already closed, I have few questions and comments:
- is this going to be implemented in 1.3.8 or a future version? I see it in the changelog but not yet associated to a version
- I think in line 4322 should be `return self.imshow(img, ax=ax, origin='upper')` and not `return basemap.imshow(img, ax=ax, origin='upper')`.
- I think the print statement in line 4320 can be removed or maybe replaced with something more robust for logging? 
Thanks for the great work

molinav:
Hi @guidocioni!

This PR was merged into the `develop` branch, so it is targeting the next minor release (1.4.0). I wanted to finalise 1.4.0 long ago but I just did not find enough spare time.

The patch releases (like 1.3.8) are branched directly from `master` (the latest stable tag) to correct things that are not working as supposed. When the fixes are done, I merge back to `master` to create a new tag and the corrections are also merged to `develop` so that they are also included in the next minor release.

If you see that something like you said is not working, feel always free to create a new PR into `develop`. One of my goals for 1.4.0 was to start having a basic test suite running, to have reports on the code coverage and prevent easy-to-detect bugs as you found.

PR #580, opened 2023-07-16
Fix AttributeError with Pillow >= 10.0

cgohlke:
The deprecated `Image.ANTIALIAS` constant was removed in Pillow 10.0. See https://pillow.readthedocs.io/en/stable/deprecations.html#constants

### DIFF
--- a/packages/basemap/src/mpl_toolkits/basemap/__init__.py
+++ b/packages/basemap/src/mpl_toolkits/basemap/__init__.py
@@ -4075,7 +4075,7 @@ class Basemap(object):
                 w, h = pilImage.size
                 width = int(np.round(w*scale))
                 height = int(np.round(h*scale))
-                pilImage = pilImage.resize((width,height),Image.ANTIALIAS)
+                pilImage = pilImage.resize((width,height),Image.LANCZOS)
             # orientation of arrays returned by pil_to_array changed in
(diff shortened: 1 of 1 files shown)


molinav:
Hi @cgohlke! Thanks for your contribution, I will try to accept the merge request as soon as possible.

I need to check if `Image.LANCZOS` is available only on recent `pillow` versions or if the name has been there for a while. Otherwise I will need to update the PR change to something like `getattr(Image, "LANCZOS", "ANTIALIAS")`.

The error in the GitHub actions is unrelated to your change, I am also trying to find a solution.

cgohlke:
`Image.LANCZOS` was added in Pillow 2.7.0, released Dec 31, 2014. See https://github.com/python-pillow/Pillow/commit/43d5d06ce4269e8ccc045f60c1bd419d67ccac4f

molinav:
@cgohlke Sorry for the late reply. Thanks for digging into the `pillow` history. This constant name is really coming from very long ago, so it should be perfectly safe to apply the PR.

As I mentioned some days before, the workflow is failing but due to an unrelated problem, which I am currently looking at. Being the PR merged into the `develop` branch, this change would appear in the `basemap` release 1.4.0.

molinav:
In the end I backported this bugfix to the `hotfix-1.3.8` branch, so it will also be available for the `basemap` release 1.3.8.

PR #593, opened 2023-11-04
Support Cython 3.0

musicinmybrain:
The main issue seems to be that the `GEOSMessageHandler` function-pointer typedef, and the callbacks with matching signatures, need `noexcept`: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#exception-values-and-noexcept


molinav (approved):

molinav:
Thanks for contributing!

musicinmybrain:
> Thanks for contributing!

You’re welcome. Thank you for doing the extra work to make my contribution fit your Cython version support needs so the PR could actually be merged.

PR #598, opened 2024-02-08
Bug fixes to arcgisimage

nitram96:
fixed bugs with using cachedir in arcgisimage.
cachedir is no longer ignored if verbose is false
if a cached image is found, now correctly shows it using self.imshow()
changed to use os.path.join to join cachedir and filename, so a backslash is no longer required at the end of cachedir

### DIFF
--- a/packages/basemap/src/mpl_toolkits/basemap/__init__.py
+++ b/packages/basemap/src/mpl_toolkits/basemap/__init__.py
@@ -4353,7 +4353,7 @@ class Basemap(object):
         ]) % (server, service, xmin, ymin, xmax, ymax, self.epsg, self.epsg, xpixels, ypixels, dpi)
 
         # Print URL in verbose mode.
-        if verbose:
+        if verbose:  # pragma: no cover
             print(basemap_url)
 
         # Try to return fast if cache is enabled.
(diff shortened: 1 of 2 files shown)


molinav:
Thanks, @nitram96! It is clear that your changes are correct, let's wait for the workflows to finish and I merge it.

Hopefully I will find some time to increase the coverage of the library. With more tests I would not have overlooked this bug.

molinav:
@nitram96 Sorry for the delay, everything looks fine, but I would like to add a couple of unit tests before merging. I hope to find some time within the next 2 days, and then most likely I will tag as `basemap` release 1.4.1 after the merge, because having such method broken in the latest stable is not nice.

molinav:
@nitram96 I took the opportunity that you opened the PR to fix a bit more the `Basemap.arcgisimage` method:

- The method code is now compliant with the most basic flake8 and pylint requirements.
- The docstring has been reformated so that it will get rendered correctly by Sphinx with the current `furo` theme.
- The cache directory should be created on the fly if saving a cache image for the first time and `cachedir` is a string corresponding to a non-existing directory. This code block was located at the wrong place, so I moved it accordingly.
- I added a couple of tests for `Basemap.arcgisimage` using the `cachedir` argument, one with an existing directory and another with a still non-existing directory.

molinav:
All seems in order, so I am merging. Thanks again, @nitram96!

PR #601, opened 2024-03-11
Allow packaging version 24.0

frenzymadness:
I'm gonna test it in Fedora Linux.


molinav:
Hi @frenzymadness! Thanks, I guess it will be no problem to update the upper pin. I usually add upper pins in all the dependencies because I had problems in the past with new releases of dependencies that were breaking the library. The downside is that these upper pins need to be updated regularly.

Let's wait until the workflows pass and I will merge it.

frenzymadness:
Thank you. I understand the reasons for the upper pins and they make sense for virtual environments. But when we update a package in Linux distribution, we have to adjust or remove them because we can have only one version of each package available (usually).

molinav:
For the future I will need to search if I can add some bot that reminds me to update upper pins whenever a new release appears in PyPI. Otherwise, in times that I am too busy I might get delayed unless people like you take the effort to create a PR.

Do you need a specific tag release in PyPI with this change or are you cloning directly from the `develop` branch?

frenzymadness:
If you have nothing else to release, don't do one just for this change. We can take it as a patch from this PR and remove it in the future when it gets invalid after the next release.

Issue #609, opened 2024-11-13
Fail to install in python >=3.12


Water-Run:
Using pip install basemap, raise exception `ModuleNotFoundError: No module named 'distutils.util'`  
Solved by `pip install setuptools`, then raise exception `AttributeError: module 'pkgutil' has no attribute 'ImpImporter'. Did you mean: 'zipimporter'?`  
Downgrade to python 3.10 to solve.

cvanelteren:
Can you checkout #611? If there are some issues I can include it in the PR

molinav:
Based on [this Stack Overflow issue](https://stackoverflow.com/questions/77364550/attributeerror-module-pkgutil-has-no-attribute-impimporter-did-you-mean), your problem appears to be the use of an outdated version of `pip` which contains symbols removed in Python 3.12, so I would expect that this is not a direct problem caused by `basemap`, but by your `pip` installation.

I would suggest that you try to upgrade `pip` and install `basemap` again. If this still does not work, please feel free to reopen the issue.

Issue #608, opened 2024-11-12
Python 3.13 compatibility


jameswilburlewis:
basemap does not "pip install" cleanly under Python 3.13.   Might be related to some of the pinned upper versions of package dependencies?  This environment had numpy 1.26.4 installed (so not a numpy 2.0 issue).

```
(pycharm_py313) jwl@jwl-new-mac pyspedas_py313 % pip install basemap       
Collecting basemap
  Using cached basemap-1.3.9.zip (156 kB)
  Installing build dependencies ... error
  error: subprocess-exited-with-error
  
  × pip subprocess to install build dependencies did not run successfully.
  │ exit code: 1
  ╰─> [63 lines of output]
      Ignoring numpy: markers 'python_version == "3.10"' don't match your environment
      Ignoring numpy: markers 'sys_platform == "darwin" and (python_version >= "3.7" and python_version <= "3.9")' don't match your environment
      Ignoring numpy: markers 'sys_platform != "darwin" and (python_version >= "3.7" and python_version <= "3.9")' don't match your environment
      Ignoring numpy: markers 'python_version == "2.7" or (python_version >= "3.4" and python_version <= "3.6")' don't match your environment
      Ignoring numpy: markers 'python_version == "2.6" or (python_version >= "3.2" and python_version <= "3.3")' don't match your environment
      Ignoring cython: markers 'python_version == "3.2"' don't match your environment
      Collecting setuptools
        Using cached setuptools-75.4.0-py3-none-any.whl.metadata (6.8 kB)
      Collecting wheel
        Using cached wheel-0.45.0-py3-none-any.whl.metadata (2.3 kB)
      Collecting numpy==1.23.3
        Using cached numpy-1.23.3.tar.gz (10.7 MB)
        Installing build dependencies: started
        Installing build dependencies: finished with status 'done'
        Getting requirements to build wheel: started
        Getting requirements to build wheel: finished with status 'error'
        error: subprocess-exited-with-error
      
        × Getting requirements to build wheel did not run successfully.
        │ exit code: 1
        ╰─> [32 lines of output]
            Traceback (most recent call last):
              File "/Users/jwl/anaconda3/envs/pycharm_py313/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
                main()
                ~~~~^^
[...]

Issue #604, opened 2024-06-17
Numpy 2.0 Compatibility 


jbpezent:
The recently released numpy 2.0 appears to be incompatible with latest basemap version 1.4.1. Seems to be a common issue with packages compiled against older versions of numpy. Issue observed on windows 11 with python 3.9 when importing basemap.

![BaseMapError](https://github.com/matplotlib/basemap/assets/40646929/8f0e650a-52a6-4066-adff-4eeec01d5d4a)

DWesl:
As I recall, the fix for this would be adding metadata to the effect of `requires: numpy<2.0` to existing binary releases and re-compiling new binary releases with `numpy>=2.0`.  

In the meantime, installing basemap from source with `pip install --force-reinstall --no-binary :all: basemap` should work until basemap does a new release with new binaries (`1.4.1post1` might be fastest, but the maintainers might be close enough to 1.4.2 to wait for that).

DWesl:
> @DWesl I can just go open a PR at feedstock with the new pin, and create a new build, in the meantime, if you OK with that? If `basemap` does indeed support `numpy>=2.0.0` at upstream (here), that is 🍺

My platform only has up to Python 3.9, so I'm not able to check anything with NumPy 2.0 until someone works out how to compile Python 3.12 there and packages it.

molinav:
My main concern while I keep the pin `numpy<2.0` for now is that NumPy 2.0 brings some API breaking changes that have been waiting around for some years. Since `basemap` was started very long ago, I would not be surprised that `basemap` has some old ways of doing things (e.g. using NumPy members that were removed or moved somewhere else).

Because of that, I would like to invest a bit of time reviewing the source code before pushing a new release with potential bugs (note that the test collection is quite limited). Unfortunately, I have been a bit busy in the last months, but I have not still forgotten the latest open issues and PR.

cvanelteren:
Any news on this?

cvanelteren:
Can you checkout #611? If there are some issues I can include it in the PR

molinav:
This issue is solved by @cvanelteren's PR #614 and it will be available when `basemap` 2.0.0 gets released soon.

cvanelteren:
The issue is solved by both of us @molinav  -- credit where credit is due;-)

rgommers:
[...]

PR #611, opened 2025-01-24
Update build system and rebuild with numpy >= 2.0

cvanelteren:
Hi there,

I made some changes to the build chain and update the numpy dependencies. The tests pass both here as in our package [Ultraplot](https://github.com/Ultraplot/ultraplot/tree/main). Let me know if you wish me to comment on these changes.
Sincerely,
C

### DIFF
--- a/.github/workflows/basemap-for-manylinux.yml
+++ b/.github/workflows/basemap-for-manylinux.yml
@@ -90,7 +90,7 @@ jobs:
     container: "pylegacy/${{ matrix.arch }}-python:3.8-debian-4"
     steps:
       - name: Download checkout
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: checkout
           path: .
@@ -145,12 +145,12 @@ jobs:
     container: "pylegacy/${{ matrix.arch }}-python:${{ matrix.python-version }}-debian-4"
     steps:
       - name: Download checkout
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: checkout
           path: .
       - name: Download GEOS artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-geos-${{ matrix.arch }}
           path: ${{ env.PKGDIR }}/extern
@@ -229,12 +229,12 @@ jobs:
     container: "pylegacy/${{ matrix.arch }}-python:3.8-debian-10"
     steps:
       - name: Download GEOS artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-geos-${{ matrix.arch }}
           path: ${{ env.PKGDIR }}/extern
       - name: Download build artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-build-${{ matrix.arch }}-${{ matrix.python-version }}
           path: ${{ env.PKGDIR }}/dist
@@ -283,12 +283,12 @@ jobs:
     container: "pylegacy/${{ matrix.arch }}-python:${{ matrix.python-version }}-debian-10"
     steps:
       - name: Download checkout
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4444
         with:
           name: checkout
           path: .
       - name: Download build artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-build-${{ matrix.arch }}-${{ matrix.python-version }}
           path: ${{ env.PKGDIR }}/dist
@@ -327,7 +327,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v1
       - name: Download test artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-test
           path: ${{ env.PKGDIR }}
@@ -353,7 +353,7 @@ jobs:
     container: "pylegacy/python:3.8-debian-10"
     steps:
       - name: Download checkout
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: checkout
           path: .
@@ -362,7 +362,7 @@ jobs:
           cd ${{ env.PKGDIR }}
           pip install -r requirements-doc.txt
       - name: Download build artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-build-x64-3.8
           path: ${{ env.PKGDIR }}/dist
@@ -424,7 +424,7 @@ jobs:
     environment: PyPI
     steps:
       - name: Download build artifacts
-        uses: actions/download-artifact@v1
+        uses: actions/download-artifact@v4
         with:
           name: artifacts-build-${{ matrix.arch }}-${{ matrix.python-version }}
           path: ${{ env.PKGDIR }}/dist
--- a/.github/workflows/basemap-for-manylinux.yml
+++ b/.github/workflows/basemap-for-manylinux.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v4
       - name: Upload checkout
         uses: actions/upload-artifact@v3
         with:
@@ -325,7 +325,7 @@ jobs:
     container: "pylegacy/python:3.8-debian-10"
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v4
       - name: Download test artifacts
         uses: actions/download-artifact@v4
         with:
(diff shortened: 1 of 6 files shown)


cvanelteren:
The PR would address #604 #608  #609

cvanelteren:
Perhaps one of the maintainers can comment on the build failure. Pinging @kant  :bell:

molinav:
@cvanelteren The error is coming because of the deprecated versions used in all the artifacts actions; `v1` was needed because of using pleistocenic base images to still build with the `manylinux1` tag.

I appreciate all your efforts in this PR, and I think it is actually a good moment to start thinking of `basemap` 2.0.0 and bring all the breaking changes that I was avoiding for the purpose of keeping the legacy things working. Some items I can list fast right now:
- Moving to "Python >= 3.X", with X to be defined (reason for 3.7? better 3.8 because of the Cython 3.1 target versions? 3.9 to stay with the non-EOL versions?
- Moving from `manylinux1` to `manylinux_2_17` for the GNU/Linux wheels. This already allows the use of not so old base images in the GitHub Actions, and probably `v4` of the artifacts actions may work *as is*.
- Start relying on implicit namespace packages (no empty `__init__.py` files). There is a PR from long ago still open.
- Migrate everything in `setup.py` and `setup.cfg` to `pyproject.toml`. Is this even possible, considering that GEOS needs to be built? I do not have the knowledge but I see that you define something about that in the `pyproject.toml` of this PR.

My only concern with the current PR is that it makes several big changes all at once, I would be in favour of addressing each of the current problems one by one in separate PR. I think that, as of today, the `develop` branch should become the branch targeting a major release `basemap` 2.0.0. I will look at the latest commit and create a `support-1.x` branch, so that I can still keep my legacy support without disturbing the majority of the `basemap` users, because it really is a problem for you right now: the latest `basemap` patch release from my side was almost 1 year ago, and there is still no support for Python 3.13 and `numpy >= 2.0`.

DWesl:
> * Migrate everything in `setup.py` and `setup.cfg` to `pyproject.toml`. Is this even possible, considering that GEOS needs to be built?

It is not currently possible to create extension modules in `pyproject.toml`, but all the other metadata should be there.  I think there's a PR over on PyPA/setuptools about what extension modules in `pyproject.toml` should look like.

cvanelteren:
@molinav no worries if this PR is trashed ;-).  I wasn't keeping track of the main. I am happy to chip in to work on stuff. I don't have full overview of the pipeline but noticed some verbose and missing stuff in `pyproject.toml` (hence this PR). 

I agree that this is perhaps a bit too rushed and fast. We could split this PR up by first tackling `pyproject.toml` (to move everything there, and then update the dependency for higher python versions (say 3.9 and up?). Let me know what you think.

To answer one of you questions:
>Migrate everything in setup.py and setup.cfg to pyproject.toml. Is this even possible, considering that GEOS needs to be built? I do not have the knowledge but I see that you define something about that in the pyproject.toml of this PR.

Yes I have included it as such in the PR here

ReimarBauer:
I'e just seen that there is a ruff checker available. 
https://numpy.org/devdocs/numpy_2_0_migration_guide.html#numpy-2-migration-guide

cvanelteren:
I have some time to work on this but don't wanna step on anyone's toes. If the team could outline what they have in mind and my time is appreciated then by all means I can and will under some guidance where you wanna move the package.

molinav:
I would really appreciate your help, @cvanelteren. Unfortunately, I am quite overloaded and close to my limits, so I doubt I can be hands-on myself at least until the second half of March. Reviewing changes should be something that I can still handle despite my limited availability.

My comment https://github.com/matplotlib/basemap/pull/611#issuecomment-2612592179 above summarises the main changes needed. The following PRs into `develop` would be sensible, in this order (and making each PR address one thing at a time):
- [ ] PR making the GitHub workflows work again. Here, we have two possibilities:
   1. Fix the current workflows. This means upgrading all the artifacts actions to v4. However, because of the `node` requirements for these actions, my Debian-based running the workflows need to be bumped to something newer (e.g. Debian 8?). This implies that the generated wheels will not be `manylinux1` anymore, but at least `manylinux_2_17`. The current workflows are limited to Windows and GNU/Linux (glibc): `musllinux` wheels are missing, and MacOS wheels were generated and uploaded manually in one of my personal machines.
   2. Reimplement the workflow using `cibuildwheel` to build the wheels. This is probably the long-term solution, because we delegate the build setup to a project that is dedicated exclusively to it. For example, the current workflow cannot generate wheels for Python 3.13 on GNU/Linux simply because I did not have time to rebuild my Debian-based images together with Python 3.13 (the last time I had some free time to play with it, Python 3.12 was the latest). Here, I cannot be of much help due to my lack of knowledge with `cibuildwheel`.

- [ ] PR removing support for some EOL Python versions. Depending on the approach for the former PR:
   1. With my Debian-based images, the minimum Python version is not really relevant, but it should be at least Python 3.3 because there is the PR https://github.com/matplotlib/basemap/pull/576 blocked for long time requesting the use of implicit namespace packages (i.e. removing empty `__init__.py` files as indicated in [PEP 420](https://peps.python.org/pep-0420/)). Probably, other newer Python version is more sensible, but still not sure which one.
   2. With the `cibuildwheel` approach, the supported Python versions are constrained to what `cibuildwheel` decides to support, I think they follow a strict version upgrade based on the Python version support, but I am not sure about the details (support for all non-EOL versions or for the four latest non-EOL versions?

- [ ] PR https://github.com/matplotlib/basemap/pull/576 can then be revisited and accepted (only the changes related to the implicit namespace packages).

- [ ] PR with the upgrade for `numpy >= 2.0`.

All these PRs (fixed workflows, `manylinux_2_17` wheels, Python 3-only migration, NumPy 2.0 support, Python 3.13 support) should be merged into `develop` to create the next major release `basemap` 2.0.0. I am happy if I can create a `support-1.x` branch from the current latest `develop` commit (https://github.com/matplotlib/basemap/commit/461b40269088c936e2feaea12a33991bd96b63b5) and continue some support for `basemap` 1.x, but I do not want that my specific requirements block the main `basemap` project. It was okay when I took over the maintenance in 2021, because the project was stale (we did not even have wheels on PyPI), but it is time to move on in the `develop` branch.

cvanelteren:
Thanks for the guidelines. I started working on the first task here: https://github.com/matplotlib/basemap/pull/612

cvanelteren:
Closing in favor of #614

cvanelteren:
I think all points above are now addressed @molinav. Can we proceed to push develop to main?

molinav:
@cvanelteren All the major changes are now addressed. I would like to mention some points regarding open items:

- [X] https://github.com/matplotlib/basemap/pull/617 may remain open for the time being. I realised that `flake8` and `pylint` were only running on a subset of files. A lot of cleanup is still needed before we can run both in the whole library. Furthermore, I would avoid free refactorings based on `black` at least until the library is clean (I am also not a big fan of `black`, but this is subjective).

- [X] https://github.com/matplotlib/basemap/issues/616 may remain open for the time being. In this issue, I see two different topics:
  - The migration to `pyproject.toml`. I always have mixed feelings with these configuration files, because in the end any Python package seems to still need a `setup.py` for the dynamic things. I would not mind moving everything possible to `pyproject.toml`, and especially to migrate the `setup.cfg` options to it. We have three files (`setup.py`, `setup.cfg` and `pyproject.toml`) to handle the metadata and build process when the legacy approach only needed one (`setup.py`). Removing `setup.cfg` at least allows to go back to two (`setup.py`, and `pyproject.toml`). But I am not sure it makes sense to block `basemap` 2.0.0 for this refactoring, when it currently works and we can address it e.g. in `basemap` 2.1.0.
  - The project restructuring. In this item I have to disagree, but from your open issue I see that the documentation on the project structure needs to be improved. We need to distinguish "project" from "package" (you must have seen the two terms already during the `cibuildwheel` migration because they use the same terms). For most projects, one single package is developed, and the concept of project and package mix. In the `basemap` project, we build *three* packages, so project and package do no overlap directly. If you move to the `packages` folder (`basemap`, `basemap_data`, `basemap_data_hires`), each subfolder corresponds to a Python package tree with the standard `src`-based package skeleton. The approach suggested in the issue also moves from a `src`-based skeleton to a flat skeleton, and I think this is not a good idea. I would like to describe my reasons better in the issue itself, but I am quite limited in time right now.

- [X] The current `cibuildwheel` workflow does not build the wheels from the sdists. I am not totally happy with it, because the sdist -> wheel approach guarantees that the sdists allow to build wheels for the platforms in which precompiled wheels are not available. I uploaded broken sdists in the past here because of not using the sdist -> wheel approach. The legacy workflows implemented the build process in this manner to address this problem. I guess we can live with it for now, but it should be fixed at some point (I was playing around with `cibuildwheel` yesterday but without success).

- [ ] The legacy workflows had a specific job to build the `basemap` documentation with Sphinx. This should come back, because I was using the artifacts of this job to later update the `gh-pages` branch without having to build the documentation manually.

- [ ] The `cibuildwheel` workflow needs to install `basemap-data` and `basemap-data-hires` from the previous job during the `basemap` test stage. Otherwise, I cannot move the `basemap-data` requirement below to `basemap-data >= 2.0, < 3.0` because `basemap-data` 2.0.0 and `basemap-data-hires` 2.0.0 packages are still not available on PyPI (they will be uploaded together with the release of `basemap` 2.0.0):
https://github.com/matplotlib/basemap/blob/7913a36cbc262f4641f0debdd578a54799f9d31c/packages/basemap/requirements.txt#L1

I think the last two points should be addressed before I prepare `basemap` 2.0.0 for release. The first three points can wait.

PR #614, opened 2025-02-18
Modernize CI build

cvanelteren:
This was rather frustrating to figure but I have gotten it to build using CI and not touching the other part of the codebase. The major issue was that some of the docs on CIBUILD were confusing with paths changing without really good documentation on it.  At least the CIBUILD is transferred now and we can move on to other things. There are some warnings that needs addressing in another PR.


cvanelteren:
I need to add the pytest but otherwise it looks good

packages/basemap/pyproject.toml
```diff
@@ -2,14 +2,7 @@
...
     'setuptools',
     'wheel',
-    'numpy == 1.26.1; python_version >= "3.12"',
-    'numpy == 1.23.3; python_version == "3.11"',
-    'numpy == 1.21.4; python_version == "3.10"',
-    'numpy == 1.21.4; sys_platform == "darwin" and (python_version >= "3.7" and python_version <= "3.9")',
-    'numpy == 1.16.6; sys_platform != "darwin" and (python_version >= "3.7" and python_version <= "3.9")',
-    'numpy == 1.16.6; python_version == "2.7" or (python_version >= "3.4" and python_version <= "3.6")',
-    'numpy == 1.11.3; python_version == "2.6" or (python_version >= "3.2" and python_version <= "3.3")',
+    'numpy >= 2.0.0; python_version >= "3.9"',
```
molinav, line 5: I suggest here `    'numpy == 2.0.0',`:
- We need `==` and not `>=` because this is `numpy` as build requirement, and we want to pin to the oldest supported `numpy` version.
- Since the PR effectively sets the minimum supported Python version to 3.9, we do not need anymore the block `; python_version >= "3.9"`.
- Consequently with the previous point, I would also suggest that you edit the `setup.py` file and set `"python_requires": ">=3.9, <3.13",`.

.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
...
+      matrix:
+        package: [basemap_data, basemap_data_hires]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.8"
```
molinav, line 34: Since the PR effectively moves the support to start with Python 3.9, we can also use Python 3.9 here (the minimum supported Python version).

.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
...
+        env:
+          CIBW_PROJECT_REQUIRES_PYTHON: ">=3.9"
+          CIBW_ARCHS_LINUX: auto64
+          CIBW_BUILD_VERBOSITY: 1
+          CIBW_BUILD: "cp39* cp310* cp311* cp312*"
+          CIBW_SKIP: "pp* *-musllinux_* *-win32 *-manylinux_i686 *-musllinux_i686 *-aarch64 *-armv7l"
+          CIBW_BEFORE_ALL: ${{ matrix.before_all }}
+          CIBW_ENVIRONMENT: ${{ matrix.environment }}
+          CIBW_BUILD_REQUIRES: "cython>=0.29,<3.1"
+          CIBW_REPAIR_WHEEL_COMMAND: ""
```
molinav, line 116: I think this `CIBW_REPAIR_WHEEL_COMMAND` needs some tuning. I was looking at the GNU/Linux wheels and they are tagged as `linux_x86_64`, so these means that they are probably not repaired. I looked inside them and the GEOS shared library is not bundled, so the wheels will not work out of the box and also will be rejected by PyPI in the event of an upload. The former workflow was using `auditwheel repair` for this, so I would expect that something similar here should work.

The missing repair step only affects the future manylinux wheels. The Windows wheels are generated correctly and they bundle `geos_c.dll`, because the bundling is done with a trick in the `setup.py` file. In the future it can be inspected if this approach may be changed to use `delvewheel`, but I would leave it for now.

.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
...
+              Write-Host 'Starting BEFORE_ALL script';
+              $pwd = Get-Location;
+              Write-Host 'Current directory:' $pwd;
+              Set-Location '{package}';
+              $env:GEOS_DIR = '{package}/extern';
+              python -c 'import sys, os; sys.path.append(\"utils\"); from GeosLibrary import GeosLibrary; GeosLibrary(\"3.6.5\").build(\"extern\", njobs=16)'
+              "
+            environment: >-
+              GEOS_DIR="D:/a/basemap/basemap/packages/basemap/extern"
+              CFLAGS="-ID:/a/basemap/basemap/packages/basemap/extern/include -ID:/a/basemap/basemap/packages/basemap/extern/include/geos/include"
```
molinav, line 87: This is just a comment for the future: I would like to see if `CFLAGS` and `LDFLAGS` can be removed from here at some point *after* this PR. Looking at the logs of the Windows builds, it can be seen that `GEOS_DIR` is identified successfully by the `setup.py` file, so `CFLAGS` and `LDFLAGS` are set correctly by `setup.py` without the need to override them manually.

.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
...
+        os: [ubuntu-latest, windows-latest]
+        include:
+          - os: ubuntu-latest
+            before_all: |
+              echo "Starting BEFORE_ALL script"
+              cd "{package}"
+              export GEOS_DIR="$(pwd)/extern"
+              echo "GEOS_DIR set to: $GEOS_DIR"
+              python -c "import sys; sys.path.append('utils'); from GeosLibrary import GeosLibrary; GeosLibrary('3.6.5', root='extern').build('extern', njobs=16)"
+            environment: >-
```
molinav, line 65: This is just a comment for the future: The logs from building on GNU/Linux show a warning that `GEOS_DIR` was not defined or does not allow to find the GEOS library. My feeling here is that the `GEOS_DIR` variable needs to be defined again in this `environment` block and it is not enough to have it in the `before_all` block. The GNU/Linux build still succeeds because `CFLAGS` and `LDFLAGS` are hardcoded to the final correct values. A task *after* this PR would be to add `GEOS_DIR` here and see if the warning at build time disappears, and later remove `CFLAGS` and `LDFLAGS` and check that everything continues working.

molinav (commented):
I added some comments to your changes. The only one that is important is the one regarding the missing `auditwheel repair` (targeting manylinux2014/manylinux_2_17), the rest is all minor.

.github/workflows/build.yml
cvanelteren: Hmm that is strange. I have looked at what the utils directory is doing but in my runs it wasn't picking it up without setting this manually.

.github/workflows/build.yml
cvanelteren: Yeah this stuff is giving me headaches. I was trying to debug when what is available. But I seem you have nicely fixed a few things I was running into too and failing to fix on a separate branch. Thanks!

molinav:
If you mean the additional data files (boundaries, rivers, etc.), they belong to `basemap-data` and `basemap-data-hires`, and these packages are generated (in sdist and wheel form) in two other artifacts by the workflow.

But now that you mentioned this, I realised that what we are missing is the sdist version of `basemap` itself. We only have precompiled wheels for `basemap`, but the source distribution also needs to be generated.

cvanelteren:
The `sdist` is added. Should the data artefacts be merged or this done in the process?

molinav:
I think it is fine for now to have separate artifacts, as long as the upload job picks all of them later.

molinav:
I want to pick one wheel for Windows and one for GNU/Linux to see that they install and import properly on my machine, and then it is time to merge!

molinav:
@cvanelteren Adding `GEOS_DIR` to the GNU/Linux environment removes the warning about GEOS not being found in the `cibuildwheel` step. I will try to remove `CFLAGS` and `LDFLAGS` now; in case this fails, I will remove the commit.

.github/workflows/build.yml
molinav: Resolving here because it was safe to remove `CFLAGS` and `LDFLAGS` after setting `GEOS_DIR`.

.github/workflows/build.yml
molinav: Resolving here because it was safe to remove `CFLAGS` and `LDFLAGS` after setting `GEOS_DIR`.

molinav:
@cvanelteren I am merging your PR now. I want to play a bit more with a couple of things, but I can make it in a separate branch in my fork without delaying your PR more. In the future I would like to:
- Improve the setup of `GEOS_DIR` (it is working but not as I would be expecting).
- Make `before_all` platform-independent. For this, a Python script in `.github/workflows` can be added doing what the `before_all` snippets are doing right now.
- At some point, linting (at least `flake8`) and testing (with `pytest`) need to be added back to the workflow.

Thanks a lot for your effort!

cvanelteren:
No worries @molinav. I can assist in this if you would like. To respond to your points

- I had some issues too with the build platforms having platform specific issues and or not working as expected. Hence the roundabout way of doing it. Moving it to a platform agnostic (read python centric) would benefit in future maintainability 
- if the current implementation works, there is no need but I agree with the previous point I wrote here
- linting could be performed with a precommit to a a PR, pytest could be added to the current build work flow at the end stage

PR #615, opened 2025-02-18
Add necessary noexcept for more recent cython

cvanelteren:


molinav:
Thanks, @cvanelteren! I also added a couple of changes that are related to the new syntax.

PR #622, opened 2025-02-26
Cibuild built from sdist -> wheel

cvanelteren:
This PR achieves two things.
First it builds the wheels directly from sdist rather than building the two separately.
Second, it ensures that cibuild will use the data and data_hires in the builds.

This was discussed on #611.


cvanelteren:
Hi @molinav does this align with your intention?

PR #623, opened 2025-02-26
Add docs to workflow

cvanelteren:
As it says on the tin. Branches from #622


cvanelteren:
Marked as draft until #622 is merged.

PR #627, opened 2025-07-18
CI: Harden GHA configuration

tacaswell:
Apply recommended hardening steps including:

- pinning to a SHA any actions used
- not persisting the read token on checkout
- setting the default permissions
- adding a depandabot file for GHA

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -119,7 +119,7 @@ jobs:
           echo "SDIST_DIR=$extractedDir" | Out-File -FilePath $env:GITHUB_ENV -Append
 
       - name: Build basemap wheels from sdist
-        uses: pypa/cibuildwheel@v2.22.0
+        uses: pypa/cibuildwheel@ee63bf16da6cddfb925f542f2c7b59ad50e93969 # v2.22.0
         env:
           CIBW_ARCHS: "native"
           CIBW_BUILD: "cp39* cp310* cp311* cp312* cp313*"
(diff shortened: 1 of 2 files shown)


tacaswell:
I apparently forgot to pull main before doing this, working on it.

molinav:
Thanks for the contribution, @tacaswell! I was looking at the workflow attempt for this PR:
https://github.com/matplotlib/basemap/actions/runs/16377453139

The workflow could not start with the following error message:
```
Invalid workflow file: .github/workflows/build.yml#L2](https://github.com/matplotlib/basemap/actions/runs/16377453139/workflow)
The workflow is not valid. .github/workflows/build.yml (Line: 2, Col: 3): Unexpected value 'content'
```

It refers to the first code block added at the top of `.github/workflows/build.yml`:
https://github.com/matplotlib/basemap/blob/93310cc070434093f4fdbe58364bfb6518883b0f/.github/workflows/build.yml#L1-L2

Something in this code block seems to need some syntax change (I do not have the knowledge for this global use of `permissions`).

molinav:
I was looking briefly at other similar merged PRs in the `matplotlib` ecosystem:
https://github.com/matplotlib/matplotview/pull/8/files
https://github.com/matplotlib/mpl-gui/pull/22/files

It seems to me it is just a typo, so it should be `contents` instead of `content`. Is this correct?

.github/workflows/build.yml
```diff
@@ -1,3 +1,5 @@
+permissions:
+  content: read
```
tacaswell:
```suggestion
  contents: read
```

molinav:
Thanks, @tacaswell! Let's merge.

